PR #114, opened 2026-06-12
fix(kalshi): settlements finally reach the ledger — venue-feed sweep + backfill CLI

DarriEy:
Description redacted — previously contained operational figures (P&L / balances / position data) not suitable for a public repo. The technical change is in the diff.

### DIFF
--- a/auramaur/bot.py
+++ b/auramaur/bot.py
@@ -3017,6 +3017,29 @@ class AuramaurBot:
                 return
             try:
                 resolved = await tracker.check_resolutions()
+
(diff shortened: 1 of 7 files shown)


DarriEy:
Verified end-to-end against production: after the lazy-init fix the sweep walks the real feed cleanly — and it's genuinely **empty**: no Kalshi position has ever settled (the book skews long-dated 2026 markets). So nothing was silently lost historically; the machinery now books settlements as they start landing, which is what the 2-4 week scorecard needs. Dry-run output: `No unbooked Kalshi settlements found.`

PR #292, opened 2026-07-18
feat(data): add lineage and information graduation

DarriEy:
## Summary
- add point-in-time ingestion, source-fetch, evidence, forecast, calibration, and resolution lineage
- add official-source shadow connectors and a conservative information-graduation ladder
- add leakage-safe edge reporting, data-contract audits, readiness health checks, and operator commands
- reconcile information trials when markets resolve

## Safety and scope
- production evidence ranking is unchanged
- shadow sources have zero influence until graduation gates pass
- paired-forecast worker is explicitly deferred pending call budgets, immutable arm manifests, scheduling, and cost attribution
- no IBKR/ETF files or large `price_history` migration are included in this PR diff
- lineage writes use unique savepoints and scoped rollback behavior
- source-health timestamps are compared as timezone-aware datetimes

## Validation
- copied v20 database migrated to v21 successfully
- SQLite integrity check: `ok`
- Ruff: passed
- pytest: `1126 passed, 1 skipped`
- `git diff --check`: clean


DarriEy:
## Review: CHANGES REQUESTED (high-effort multi-agent review, 32 agents, findings independently verified)

Second submission of this framework; both previously-rejected regression classes are still present, and the PR description's safety claims do not survive contact with the code.

**Blocking (all CONFIRMED):**

1. **"Production ranking unchanged" is false.** New short SOURCE_AUTHORITY keys (nws/bls/bea/congress/eia, weight 3.0) are substring-matched against every item URL — a URL containing "gdp-beats-forecast" or "congress-vote" jumps to authority 3.0. `authority()` feeds the production ranker and evidence compressor, so evidence ordering changes for ALL existing strategies. Fix: exact-domain matching on parsed hostnames; shadow-source authority in its OWN map.

2. **Full-connection rollback, round two — via the commit-failure path.** Aggregator lineage: RELEASE SAVEPOINT → commit(); when commit raises (nightly lock bursts here exceed busy_timeout), ROLLBACK TO SAVEPOINT fails ("no such savepoint") and the fallback full `db.rollback()` discards concurrent tasks' uncommitted writes — including live fills. Same pattern in engine.py's new forecast_snapshots block, which also runs unconditionally for every analyzed market, is unguarded (a raising rollback masks the original exception), and aborts the analysis it rode in on.

3. **Schema version collision**: PR stamps v21; main and the production DB are already at 21 (#289). The migration will never run. Rebase to v22 and verify the skipped-migration path.

4. **Observability that lies in both directions**: readiness PASSes ("0 active") during a total ingestion outage once source_fetches has any historical row; FAILs on zero-influence shadow sources while full-scanning an unbounded table; the stuck_ingestion contract can never fire; the lineage audit flags healthy description-only forecasts.

5. **Personal email hardcoded into a public repo** — remove immediately regardless of the rest.

6. edge_vs_market locks historical calibration data out of the tool.

**Architectural requirement for round three:** the lineage system must be OBSERVER-ONLY with respect to shared infrastructure — its own DB connection (or a write queue), zero savepoints on the shared connection, zero writes inside analyze/fill hot paths, zero mutations to shared maps. Two submissions have failed on exactly these seams; the third must be structurally incapable of touching them.

Worth keeping: the lineage data model, the information-graduation concept, the deferred-worker restraint, the test volume.

DarriEy:
Addressed in `a5692a3` with the requested architectural boundary.

1. **Ranking isolation:** removed official candidate keys from the production authority map. URL boosts now parse hostnames and match exact domains/subdomains only; shadow authority lives in `official.py`. Added hostile path/hostname regressions.
2. **Observer-only lineage:** removed lineage savepoints and writes from aggregator/engine/calibration shared connections. Producers now only `put_nowait` immutable events; a background `LineageObserver` owns a separate connection, transaction, rollback domain, and fail-open logging. Forecast/resolution failures cannot abort analysis or roll back trading writes.
3. **Schema:** bumped to v22 with an explicit v21→v22 migration. Added coverage starting from a DB already stamped v21, plus copied-DB migration verification.
4. **Observability:** readiness uses a bounded production-only fetch query, does not PASS an empty window, and ignores zero-influence shadow sources. Audits now check persisted incomplete runs and dangling evidence references rather than impossible/false-positive proxies.
5. **Privacy:** removed the personal email from the NWS user agent.
6. **Historical edge:** leakage-safe snapshot reporting now supplements, rather than excludes, legacy resolved calibration rows using the last market signal at or before resolution.

Validation after the rewrite:
- copied DB: v20 → v21 → v22, `PRAGMA integrity_check = ok`
- Ruff: passed
- pytest: `1129 passed, 1 skipped`
- `git diff --check`: clean

DarriEy:
Re-review of a5692a3: all six remediation claims verified directly against the code — exact-hostname authority matching with shadow weights out of the shared map, observer-only lineage on its own connection (producers put_nowait only; private rollback domain), zero savepoints on the shared connection, schema v22 with a from-v21 migration test matching the production DB state, readiness fixes, email removed. 1,129 tests pass on the production venv. One non-blocking nit: the observer queue is unbounded — if the observer wedges, events accrue in memory; consider a maxsize with drop-and-count. APPROVED.

PR #297, opened 2026-07-18
fix(ibkr): harden multi-asset paper validation

DarriEy:
## Summary
- pin held futures/options/bonds to their original IBKR contract IDs for quotes and history
- use IBKR liquid-hours metadata for holiday, split-session, and overnight calendar checks
- make full-universe preflight bounded, pacing-aware, retryable, book-scoped, and safe alongside the daemon
- normalize IBKR qualification multipliers separately from paper P&L units (including CBOT grains)
- quarantine unsubscribed instruments explicitly while preserving them in the catalog
- qualify forward-paper edge using net P&L after commissions and financing

## Real TWS validation
- global ETFs: 20/20 contracts/history ready
- FX: 10/10 ready
- futures: 9/9 ready
- international equities: 8/8 entitled listings ready; 7203.T explicitly TSE-entitlement gated
- options: 10/10 contracts/history diagnostics ready; synthetic estimates remain non-executable until native OPRA BBOs are available
- direct bonds: 5/5 Treasury/corporate contracts ready
- all connections read-only; daemon client 3 and preflight client 97 are isolated

## Validation
- 1161 passed, 1 skipped
- ruff check .
- git diff --check

### DIFF
--- a/auramaur/cli/diagnostics.py
+++ b/auramaur/cli/diagnostics.py
@@ -156,18 +156,23 @@ def ibkr_etf_preflight():
 
 
 @main.command("ibkr-multiasset-preflight")
-def ibkr_multiasset_preflight():
+@click.option("--book", "book_names", multiple=True,
+              type=click.Choice(("global_etf", "fx", "futures",
+                                 "international_equity", "options", "bonds")))
+def ibkr_multiasset_preflight(book_names):
     """Verify contract resolution, data, isolation, and schema for six books."""
     import sys
 
     async def _run() -> int:
         from auramaur.monitoring.ibkr_multiasset_preflight import preflight
+        from auramaur.exchange.ibkr_instruments import IBKRBook
 
         settings = Settings()
         db = Database()
         await db.connect()
         try:
-            report = await preflight(settings, db)
+            books = tuple(IBKRBook(name) for name in book_names) or None
+            report = await preflight(settings, db, books=books)
         finally:
             await db.close()
         colour = {"OK": "green", "WARN": "yellow", "BLOCK": "red"}
--- a/auramaur/db/database.py
+++ b/auramaur/db/database.py
@@ -109,6 +109,8 @@ class Database:
             await self._migrate_v22_to_v23()
         if from_version < 24:
             await self._migrate_v23_to_v24()
+        if from_version < 25:
+            await self._migrate_v24_to_v25()
 
     async def _migrate_v1_to_v2(self) -> None:
         """Add category to calibration, add new tables."""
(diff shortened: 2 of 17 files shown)


DarriEy:
Remediation update (82c0266):

- rejects frozen/delayed/unknown IBKR market-data provenance; only `ibkr_live` can fill
- detects dead TWS sockets and reconnects under the existing connection lock
- makes v23→v24 migration fail closed, verifies columns before stamping, and adds v25 persisted instrument identity
- manages/exits disabled or catalog-orphaned held positions from persisted specifications
- executes hard stops/take-profits independently of spread, FX refresh, and history availability
- excludes commission/financing rows from trade counts and win rate while retaining them in net P&L
- keeps tracked multi-asset enablement default OFF

Regression coverage includes delayed quote rejection, marketDataType provenance, reconnect, v23 migration/busy failure, disabled-position exit, stop during dislocation, and fee-safe monitoring. Validation: ruff clean; full non-e2e/non-integration suite 1171 passed, 1 skipped; focused post-final-change suite 34 passed.

DarriEy:
Operational verification (2026-07-17 21:15–21:17 America/Edmonton): ran the branch against the real TWS read-only socket on port 7497 using dedicated client ID 97 and a disposable `/tmp` database. Preflight returned READY. Contract/history coverage: global ETF 20/20, FX 10/10, futures 9/9, international equities 8/8 entitled instruments (`7203.T` quarantined), options 10/10 contract/history diagnostics, bonds 5/5. All venues were closed, so executable live BBO provenance was not asserted in this run; OPRA subscription errors were observed and the options execution gate remains fail-closed as documented. No production DB or runtime switch was touched. Post-doc CI: lint and safety gate both pass.

DarriEy:
High-effort multi-agent review (independent finders + adversarial verification) of the remediation against issue #299. The hardening is real progress on items 2, 3, 6, and the quarantine/pinning work — but four confirmed defects block approval, three of them introduced by this PR. All were verified against the branch code with line-level evidence.

**1. The `ibkr_live` freshness gate bricks every book on a delayed-data-only account (items 1 and 5 fail together).** `_quote_fresh` requires `source == "ibkr_live"` exactly, and the gate sits *before* the held-position mark/exit branch (`ibkr_multiasset_paper.py` ~227). With `market_data_type: 3` (the configured default, and the only entitlement this environment has — live requests return IBKR error 10089), every quote maps to `ibkr_delayed`, so: no position is ever marked, stop-loss/take-profit never evaluate, `unmarked_positions` never drains, and the line-267 entry gate blocks everything. The same check in `ibkr_multiasset_preflight.py` makes every open-venue probe return "non-executable ibkr_delayed quote" → every book is BLOCK → the documented enablement path can never pass. The review brief was explicit: an all-delayed world must be handled sanely, not rejected wholesale. Suggested shape: treat delayed quotes as valid for *paper* marks/exits and entries with provenance recorded (`price_source='ibkr_delayed'`), and reserve the live-only requirement for any future live-order path; or gate on a per-book `allow_delayed` config that preflight honors symmetrically. Either way, protective exits must run on the best quote available, whatever its provenance.

**2. The catalog renames orphan pre-existing positions and re-introduce item 4.** The PR renames `SIL`→`SIC` and `IG_CORP_5Y`→`CORP_5Y`, while the v24→v25 migration adds `instrument_spec_json` with `DEFAULT ''` and no backfill. A pre-existing position under a renamed key has no `BY_KEY` hit and an empty spec → `_position_spec` returns None → never marked, never exited, permanently in `unmarked_positions` → its whole book is blocked from new entries. Fix: migration-time backfill of `instrument_spec_json` from the *pre-rename* catalog (including the renamed keys), plus a test that opens a position under an old key, migrates, and asserts it still marks and can exit.

**3. Per-instrument `is_market_open` skips held positions before they can be marked — non-overlapping sessions freeze a book (new defect).** The `continue` at ~221 runs before the held branch, so a held Tokyo name keeps its key in `unmarked_positions` throughout all Western sessions, and line 267 then rejects every entry candidate. An international-equity book holding names in two non-overlapping venues can *never* enter. Held positions on closed venues should be excluded from the entry-blocking set (or marked from the last session's data), with the venue-closed count surfaced in the cycle log.

**4. Item 8 (cycle-log convention) is still not met.** `ibkr_multiasset.cycle` is emitted at a single site that three `return 0` paths precede (disabled / market-closed / empty-universe), and the log carries no funnel counts. Every defect above presents as `entries=0` or silence — indistinguishable from "no signal". Per the repo convention: emit `<pillar>.cycle` unconditionally, with counts for fresh-rejected, spread-rejected, venue-closed, orphaned, and book_full.

**Also worth a test (verified plausible, not confirmed):** the liquid-hours parser (`ibkr_market_data.py` ~311) `strptime`s both ends with `"%Y%m%d:%H%M"` inside `except ValueError: continue` — the undated-end session format (`20090507:0700-1830,1830-2330`) would fail silently and mis-report `is_market_open`. Please add fixture tests for the real liquidHours string formats.

One candidate finding was refuted in verification (the `error 162` pacing-marker concern — ib_async never renders that substring), so it is not being requested; but note that this also means the `"error 162"` marker in `_is_pacing_error` is dead code as written.

Verification coverage note: a few verifier agents were lost to a session cap mid-run; the four confirmed defects above all carry direct line-level evidence and stand on their own.

PR #301, opened 2026-07-18
fix(kraken): tolerate transient NLP cache locks

DarriEy:
## Summary
- retry transient SQLite lock/busy failures for idempotent NLP cache reads and upserts
- roll back failed cache transactions before bounded backoff retries
- keep valid LLM analysis results when cache persistence remains unavailable
- continue surfacing non-lock database failures

## Why
Kraken LLM analysis could complete successfully and then be discarded when the final NLP cache write hit SQLite contention, causing missing signals and repeated LLM calls on later cycles.

## Validation
- 20 focused Kraken/cache tests passed
- full repository gate before branch split: 1175 passed, 1 skipped
- Ruff clean
- diff check clean

### DIFF
--- a/auramaur/nlp/cache.py
+++ b/auramaur/nlp/cache.py
@@ -2,9 +2,11 @@
 
 from __future__ import annotations
 
+import asyncio
 import hashlib
 import json
 
+import aiosqlite
 import structlog
 
 from auramaur.db.database import Database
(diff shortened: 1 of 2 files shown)


DarriEy:
Reviewed. The retry shape is right and the headline fix is correct — a completed LLM analysis must survive a failed cache write, and `put()` returning without raising achieves that. Two changes requested before merge:

**1. `_rollback_after_lock` can destroy a concurrent writer's transaction.** The bot shares one aiosqlite connection across coroutines, and aiosqlite interleaves queued operations — so when the cache write hits "database is locked" (the lock holder is a *different* connection/process), another coroutine's multi-statement transaction (e.g. a fill record) may be open on *our* connection at that moment. An unconditional `rollback()` aborts it mid-flight — turning a transient cache miss into lost trade data. A failed single `INSERT OR REPLACE` doesn't leave our connection in a transaction that needs manual rollback. Please drop the rollback, or at minimum guard it with `self._db.db.in_transaction` *and* evidence that the transaction is ours (which the cache path can't establish — hence: drop it). A test that opens a transaction from a sibling coroutine and asserts it survives a cache-write retry would pin this.

**2. `"database is busy"` is not a SQLite message.** SQLITE_BUSY renders as "database is locked" and SQLITE_LOCKED as "database table is locked". The second marker matches nothing; if the intent is to cover SQLITE_LOCKED, match that string — otherwise it's dead code that suggests coverage it doesn't have.

Also noting: `auramaur/risk/graduation.py` and `auramaur/strategy/engine.py` are being edited in the production checkout right now and are not part of this PR — please move that work (and this branch) to a worktree per the repo process rule; the launchd supervisor restarts the bot from this directory, so uncommitted risk-layer edits become live code on any crash-restart.

DarriEy:
Re-reviewed 315cd3f: rollback removed, real SQLite lock messages matched, and the sibling-transaction regression test pins the invariant against a real connection. Both requested changes addressed — merging.

PR #309, opened 2026-07-19
Improve IBKR paper trading risk model

DarriEy:
## Summary
- replace raw cross-asset momentum with volatility-normalized multi-horizon signals
- size positions by loss at an immutable volatility stop, with FX/multiplier-aware risk and adverse paper slippage
- gate correlated asset-class and ETF portfolio risk, and count unrealized ETF losses before opening new exposure
- add schema migration v27 for persisted stop and initial-risk data
- add a pre-registered 200-observation/180-day statistical evidence contract
- fail closed for options and bonds until Greeks/assignment and duration/accrual accounting are implemented

## Verification
- 1213 passed, 2 skipped
- Ruff clean
- git diff --check clean

## Notes
The full suite was run on Windows with a process-local fcntl compatibility stub; production POSIX locking code is unchanged.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -83,6 +83,8 @@ IBKR_QUOTE_ENVIRONMENT=live
 IBKR_QUOTE_PORT=4002
 IBKR_ENABLED=true
 IBKR_MULTIASSET_PAPER_ENABLED=true
+# 1=live. Multi-asset paper fills fail closed for 2/3/4.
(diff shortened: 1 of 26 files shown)


DarriEy:
Superseded by #310: same delta rebased onto current main (this branch double-included already-merged #308 and was unmergeable), plus a migration-test fix. Full-suite verification recorded there.

PR #314, opened 2026-07-19
fix(kalshi): harden execution and settlement accounting

DarriEy:
## Summary
- migrate Kalshi execution, order reconciliation, and settlements to current fixed-point API contracts
- make confirmed fills authoritative, preventing live pending orders from creating phantom positions
- reconcile complete paginated position snapshots and book settlement fees with fail-loud schema validation
- validate paper and live entries against fresh executable depth and record decision-time execution telemetry
- preserve subpenny/fractional behavior through entry, HTTP submission, and Kalshi exit paths
- document the hardened accounting and experiment-graduation workflow

## Safety
- paper trading remains the default
- the three live gates, kill switch, risk gateway, and absolute exposure cap are unchanged
- no live API calls or orders were made during implementation

## Verification
- full suite: 1225 passed, 2 skipped
- post-rebase core Kalshi suite: 58 passed
- git diff --check passed

### DIFF
--- a/auramaur/bot_exits.py
+++ b/auramaur/bot_exits.py
@@ -321,7 +321,8 @@ class ExitExecutionMixin:
 
         # Never sell more than we hold
         order.size = min(order.size, pos.size)
-        if order.size < 1:
+        minimum = 0.01 if market.fractional_trading_enabled else 1.0
+        if order.size < minimum:
             return False
 
         order.source = "exit"
--- a/auramaur/broker/execution_gateway.py
+++ b/auramaur/broker/execution_gateway.py
@@ -280,6 +280,9 @@ class ExecutionGateway:
         try:
             if router:
                 order = await router.route(signal, market, size_dollars, is_live)
+            elif callable(getattr(type(exchange), "prepare_executable_order", None)):
+                order = await exchange.prepare_executable_order(
+                    signal, market, size_dollars, is_live)
             else:
                 order = exchange.prepare_order(signal, market, size_dollars, is_live)
         except UnmarketableSignal as skip:
--- a/auramaur/broker/kalshi_settlements.py
+++ b/auramaur/broker/kalshi_settlements.py
@@ -34,6 +34,15 @@ log = structlog.get_logger()
 _MAX_PAGES = 25  # 200/page → up to 5000 settlements; far beyond current history
 
 
+def _field(obj, *names, default=None):
+    """Read the current fixed-point field first, with legacy compatibility."""
+    for name in names:
+        value = obj.get(name) if isinstance(obj, dict) else getattr(obj, name, None)
+        if value not in (None, ""):
+            return value
+    return default
+
+
 async def sweep_kalshi_settlements(
     db: Database, kalshi_client, *, dry_run: bool = False,
     max_pages: int = _MAX_PAGES,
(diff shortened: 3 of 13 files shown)


DarriEy (commented):
## Review — verified against the live API and current main

**Verdict: sound and genuinely needed — approve with two follow-ups requested.** I verified the load-bearing claims empirically rather than from the diff alone.

### Confirmed correct (with evidence)
- **The premise is real and current**: a live Kalshi market sampled today carries `price_level_structure: "deci_cent"` with `step: "0.0010"` — the old `round(price, 2)` clamp genuinely mispriced these markets. The `Decimal`-based per-range quantization (BUY up / SELL down) is the right shape.
- **The host migration works**: `external-api.kalshi.com` returns byte-identical responses to `api.elections.kalshi.com` (markets + exchange status probed today).
- `OrderBook.fill_to_size` exists on main; the NO-book complement construction and the `book_side`/`outcome_side` inverse mapping in `reconcile_open_orders` are consistent with `_v2_book_side` semantics (all four token x side cases check out).
- Settlement math: fee deduction, venue-cost recovery fallback, fail-loud schema guard, idempotent `source_ref` — all good. Position pagination with cursor loop is correct.
- The `pending`-order gating matches the existing single-booking design (gateway comments: resting fills are recorded once, by the monitor).
- Body stringification (`f"{count:.2f}"`, `f"{yes_price:.4f}"`) matches the API's string-dollar fixed-point convention.
- CI green; the new tests cover the four core behaviors (fixed-point settlement parsing, subpenny submission survival, tick quantization, depth capping).

### Requests before/at merge
1. **Document the production host change.** Swapping the trade-API endpoint is the single riskiest line in the PR and appears nowhere in the PR body or docs. Please add it to `KALSHI_HARDENING.md` (and note the demo host is unchanged). Consider making the host a config value rather than a literal.
2. **Answer the late-fill pillar question.** With `res.status != "pending"` gating, a resting live order that fills later gets booked by the order monitor into shared portfolio/cost-basis — but the pillar''s own `_record_position` never runs for it. If pillar exit logic depends on pillar-local tracking, late-filled positions fall through to generic exits only. If that''s intended, a one-line comment at each gate would prevent the next reader from "fixing" it; if not, the monitor needs a callback.

### Smaller notes (non-blocking)
- `place_order` no longer coerces integral counts (`int(order.size)` → `round(size, 2)`). Entry paths guard via `fractional_trading_enabled`, but exit paths can carry legacy fractional held sizes into non-fractional markets → `count: "2.50"` would be API-rejected. Suggest coercing in `place_order` when the market is non-fractional, as defense in depth.
- `price_ranges: list[dict[str, str]]` validates against today''s payloads (verified strings), but numeric drift from Kalshi would fail Market validation → market silently dropped via `parse_error`. `list[dict]` or a lax value type would degrade more gracefully.
- The fresh-book depth check applies only to `router=None` gateways (the three pillars). The main engine''s routed path bypasses `prepare_executable_order` — fine if intentional (router has its own crossing checks), but worth one sentence in the doc since the execution-samples telemetry will only ever contain pillar entries.
- `_throttle_read` also (correctly) throttles writes — consider renaming to `_throttle` before the name misleads someone into adding an unthrottled write path.

PR #315, opened 2026-07-19
Harden venue research and execution evaluation

DarriEy:
## Summary

- make Kraken validate-only positions restart-safe, strategy-isolated, and cost-basis consistent
- add executable bid/ask paper fills, unique client order IDs, API-key preflight, and research CLI tooling
- add portfolio-faithful Kraken evaluation with equal-dollar benchmarks, walk-forward warm-up isolation, beta-adjusted residuals, ATR trailing exits, stressed-cost/holdout graduation gates, and event/order-flow inputs
- add shared walk-forward and Polymarket research infrastructure, immutable decision snapshots, CLV/rebate accounting, and execution-quality hooks
- document Kraken and Polymarket research/graduation contracts

## Review follow-up

- make the public Polymarket frontend geoblock probe advisory by default; operators can explicitly select `execution.polymarket_geoblock_mode: enforce`
- refresh eligibility telemetry on a 300-second TTL, retry transient failures, and always bypass the probe for explicit exits
- keep authenticated CLOB geographic rejections as real live rejections rather than fabricating paper fills
- restore the established portfolio-wide Kelly default of 30%; this PR no longer changes global sizing
- rename the graduation evidence threshold to `min_markets` and document that the enforced 100-independent-market, positive-LCB bar intentionally may demote cells that passed the former 20-market/positive-total contract
- document decision-capture commit behavior and add direct schema 27→29 migration coverage

## Safety

- Kraken directional trading remains paper-forced by default
- Polymarket frontend eligibility is telemetry unless enforcement is explicitly configured; exits remain attemptable in either mode
- the stricter graduation contract remains enforced intentionally; structural/two-sided exempt strategies are unchanged
- live and paper position/peak state are isolated
- no new live strategy gate is enabled by this change

## Verification

- `python -m ruff check .`: passed
- focused review regression suite: **71 passed**
- final geoblock regression suite: **19 passed**
- full suite: **1244 passed, 2 skipped**
- `git diff --check`: passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ KILL_SWITCH
 venv/
 .pytest_cache/
 .ruff_cache/
+.isolated/
+.worktrees/
+.pytest-*/
 *.log
 *.log.*
 .DS_Store
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,7 +32,7 @@ When making git commits, use `Assisted-by: Claude (Anthropic)` in the commit mes
 - Daily loss limit: $200
 - Max open positions: 500
 - Minimum edge: 5% after fees
-- Kelly fraction: 40%
+- Kelly fraction: 30%
 - Confidence floor: LOW
 - Category exposure cap: 60%
 - Second opinion divergence max: 0.25
--- a/auramaur/backtest/engine.py
+++ b/auramaur/backtest/engine.py
@@ -10,7 +10,7 @@ import structlog
 
 from auramaur.db.database import Database
 from auramaur.risk.kelly import KellySizer
-from auramaur.strategy.signals import POLYMARKET_FEE_PCT
+from auramaur.strategy.signals import taker_fee_rate
 from config.settings import Settings
 
 log = structlog.get_logger()
@@ -25,6 +25,8 @@ class BacktestResult:
     max_drawdown_pct: float = 0.0
     sharpe_ratio: float = 0.0
     brier_score: float = 0.0
+    market_brier_score: float = 0.0
+    brier_improvement: float = 0.0
     accuracy: float = 0.0  # % of directional calls correct
     avg_edge: float = 0.0
     best_trade: float = 0.0
@@ -79,7 +81,9 @@ class BacktestEngine:
         edge_threshold = min_edge_pct if min_edge_pct is not None else self.settings.risk.min_edge_pct
         kf = kelly_fraction if kelly_fraction is not None else self.settings.kelly.fraction
         bank = bankroll if bankroll is not None else self.settings.execution.paper_initial_balance
-        stake_cap = max_stake if max_stake is not None else self.settings.risk.max_stake_per_market
+        configured_cap = max_stake if max_stake is not None else self.settings.risk.max_stake_per_market
+        stake_cap = bank * configured_cap if configured_cap <= 1.0 else configured_cap
+        stake_cap = min(stake_cap, self.settings.risk.max_stake_abs_ceiling)
 
         sizer = KellySizer(fraction=kf)
 
@@ -100,7 +104,8 @@ class BacktestEngine:
                 c.predicted_prob,
                 c.resolved_at,
                 c.category,
-                m.question
+                m.question,
+                COALESCE(m.exchange, s.exchange, 'polymarket') AS exchange
             FROM signals s
             INNER JOIN calibration c ON s.market_id = c.market_id
             LEFT JOIN markets m ON s.market_id = m.id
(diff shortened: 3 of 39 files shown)


DarriEy (commented):
## Review — DO NOT MERGE YET: one operational blocker, two unstated live-behavior changes

The research infrastructure here is genuinely good work — restart-safe validate-only Kraken positions with strategy isolation, live/paper peak separation, stressed executable shadow fills, immutable decision snapshots, the authenticated user stream, and a statistically honest LCB graduation test. Tests are added where they matter (including the new fail-closed triple-gate test). But three things must be resolved consciously before this merges, because I verified them empirically:

### 1. BLOCKER: the geoblock preflight will halt live Polymarket on this deployment — including exits
I probed the endpoint this PR gates on. From the production host, `https://polymarket.com/api/geoblock` returns **`{"blocked": true, "country": "CA", "region": "AB"}`** — today, right now — while the bot is simultaneously filling live Polymarket orders through the CLOB API. Two consequences:

- **Merging this makes `_check_geoblock` reject every live order for the session** — and because exits also flow through `place_order`, the ~174 open live positions become unmanageable (stop-loss/take-profit dead). Whatever the entries decision, **exits must be attemptable**: let the venue reject them if it truly wants to; do not strand held inventory behind a local preflight.
- The frontend endpoint and CLOB API eligibility are evidently **not the same predicate** (blocked=true vs. orders filling). Gating live execution on an unofficial frontend endpoint conflates web-app geofencing with API permission.

Also: a transient network failure of that one endpoint sets `_geoblocked = True` cached **for process lifetime** — a single DNS blip silently converts the whole session to rejected live orders. If the check stays, it needs a TTL/retry and an exit-path bypass.

Separate from the code: the endpoint response itself is information the operator should weigh deliberately — that is a compliance judgment, not something a hardening PR should decide as a side effect.

### 2. Unstated: global Kelly fraction 0.30 → 0.10
This cuts **every** live position size to a third, across all venues, and appears nowhere in the PR body. Defensible on the merits ("forecast error dominates theoretical Kelly"), but it is a portfolio-wide sizing change riding in a research PR. Split it out, or state it in the summary and ratify it knowingly.

### 3. Unstated: the new graduation bar demotes the currently-live cells on deploy
`min_events` 20 → 100 **and** reinterpreted from ledger events to independent markets, plus the positive-LCB requirement. Applied to the current ledger, the only cells that have earned live capital — `resolution_lens × weather` (~20+ events) and `bias_harvest × other` (27 events) — fall far below the new bar and flip back to paper-forced the moment this deploys. Maybe that is the intended standard going forward — but "graduation remains fail-closed" undersells "your two proven live edges stop trading live." Options: state it and accept; grandfather existing probation cells; or run the new bar in observe alongside the old enforce for a window before switching.

### Smaller notes (non-blocking)
- `min_events` now means "independent markets" — rename (`min_markets`) before the old meaning bites someone.
- `lower_bound()` returns -inf below n=2, so a single-market windfall can never graduate — good — but worth a comment since it looks like a bug at first read.
- The decision-capture DB write sits on the entry hot path; it is guarded, but consider batching if entry latency ever matters.
- Schema 27 → 29 in one PR: confirm the migrations test covers 27→29 directly.
- CLAUDE.md still documents Kelly at 40% while the tracked default was 0.30 and is now 0.10 — the doc needs to follow whichever number survives this review.

Happy to re-review quickly once the geoblock exit-path and the two disclosures are addressed.

DarriEy:
Addressed all review findings in f9e76cf:

- frontend geoblock telemetry is advisory by default, TTL-bound, and retryable
- explicit exits bypass the frontend probe even in enforce mode
- authenticated CLOB geoblock failures stay live rejections (no paper-fill fallback)
- reverted the unrelated Kelly reduction; tracked default remains 30%
- renamed min_events to min_markets, documented the intentional stricter graduation rollout and n<2 LCB behavior
- documented decision-capture durability/batching tradeoff
- added direct v27→v29 migration coverage

Verification: ruff clean; focused suites green; full suite 1244 passed, 2 skipped.

DarriEy (commented):
## Re-review — all concerns resolved; this has my approval

Verified in the updated diff (not just the PR body):

1. **Geoblock**: advisory by default with `enforce` as an explicit operator opt-in; 300s TTL with transient failures uncached (retried on the next order); **exits always bypass the probe**; authenticated CLOB rejections stay honest live rejections. This is exactly the right shape — telemetry now, policy only when deliberately chosen.
2. **Kelly**: the global sizing change is fully reverted; the PR now only corrects the stale CLAUDE.md figure to the real 30%.
3. **Graduation**: `min_markets` rename done throughout, and the demote-on-deploy consequence for the current probation cells is disclosed as intentional in the Safety section — a ratified decision, not a side effect.
4. Migrations run 27→28→29 stepwise with direct test coverage; decision-capture behavior documented.

CI green. Merge when ready. One deploy note for the operator: on rollout, expect `resolution_lens × weather` and `bias_harvest × other` to flip to paper-forced under the new bar (as disclosed), and watch the first live entry cycle for the `polymarket.geoblock_advisory` telemetry line — it will fire, since the frontend probe currently reports this host blocked.

PR #317, opened 2026-07-19
Harden IBKR research evaluation and controls

DarriEy:
## Summary
- make IBKR, ETF paper, and multi-asset paper experiments explicit opt-ins across tracked config, Compose, and operator docs
- add schema v30 cost-adjusted IBKR round trips with entry/exit fill lineage and use the strict 200-observation / 180-day evidence contract in preflight
- add a deterministic momentum ETF control plus cash, buy-and-hold, and seeded-placebo comparisons
- add purged/embargoed walk-forward evaluation and portfolio metrics
- add execution-free research primitives for mean reversion, relative value, futures carry+trend, and FX carry+trend

## Safety
- IBKR remains disabled by default
- multi-asset and ETF experiments remain paper-only and opt-in
- advanced signals expose no broker, sizing, or execution surface
- options and bonds remain disabled

## Verification
- combined IBKR suite: 121 passed
- full suite: 1267 passed, 2 skipped
- `python -m ruff check .`
- `git diff --check`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -77,12 +77,12 @@ AURAMAUR_LIVE=false
 # arm them explicitly on the final host only after verification.
 AURAMAUR_CONTAINER_LIVE=false
 AURAMAUR_CONTAINER_ENABLE_TRANSFERS=false
-# The dedicated IBKR live login is a read-only quote source. Port 4002 is an
-# intentional deployment-local choice; local paper execution remains separate.
-IBKR_QUOTE_ENVIRONMENT=live
+# IBKR is opt-in. Start with a paper Gateway/login; the socket stays read-only
+# and strategy fills remain in local paper ledgers.
+IBKR_QUOTE_ENVIRONMENT=paper
 IBKR_QUOTE_PORT=4002
-IBKR_ENABLED=true
-IBKR_MULTIASSET_PAPER_ENABLED=true
+IBKR_ENABLED=false
+IBKR_MULTIASSET_PAPER_ENABLED=false
 # 1=live. Multi-asset paper fills fail closed for 2/3/4.
 IBKR_MARKET_DATA_TYPE=1
 # Stable private addresses make the Gateway Trusted IP entry survive restarts.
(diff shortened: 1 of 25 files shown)


DarriEy (commented):
## Review — approve; one deployment action required at rollout

This is the right shape for the IBKR program: explicit opt-in everywhere, cost-adjusted evidence, and honest controls. Verified points:

- **Opt-in flip is coherent across all three layers** (tracked config `enabled: false`, Compose env defaults `false`/`paper`, `.env.example` documenting the knobs), with a new `test_ibkr_default_safety` regression pinning it. A fresh checkout can no longer start an experiment just because a Gateway is reachable — good.
- **Schema v30 round trips**: entry/exit fill lineage with `exit_fill_ref UNIQUE` for idempotency, net P&L subtracting both commissions; the cost columns for financing/borrow/roll/intelligence exist but are zero-filled for now, which is fine as a forward contract.
- **Preflight now consumes `evaluate_ibkr_evidence`** over cost-adjusted round trips (200-obs/180-day contract) instead of the old 30-exit profit-factor heuristic — a real upgrade, and the "graduated / not graduated + reasons" wording is operator-friendly.
- Deterministic momentum control + cash/buy-and-hold/seeded-placebo comparisons and purged/embargoed walk-forward are exactly the anti-self-deception tooling the evidence contract needs.
- CI green (full suite), 121 IBKR tests passing.

### Required at deploy (operational, not a code change)
The Compose defaults flip means the production deployment goes **IBKR-dark on rollout unless `.env` pins the knobs**: env vars beat the local YAML override, so the existing `defaults.local.yaml` `ibkr.enabled: true` will be overridden by `IBKR__ENABLED=false` from the new Compose default. Before (or with) the deploy, the production `.env` needs:

```
IBKR_ENABLED=true
IBKR_QUOTE_ENVIRONMENT=live
```

…and later `IBKR_MULTIASSET_PAPER_ENABLED=true` when the multiasset review-hold is lifted. Without these, the quote connection and this week''s market-data-entitlement work land on a dormant integration and the registry never revalidates.

### Minor notes (non-blocking)
- The round-trip INSERT is unguarded against a crash-replayed exit (UNIQUE violation would fail the cycle once); a `INSERT OR IGNORE` or try/except would make replay silent.
- `opened_at` falls back to `now()` when the position row is missing — that understates `elapsed_days` for the evidence contract in the degenerate case; worth a debug log when the fallback fires.

PR #322, opened 2026-07-20
Web dashboard, Kalshi accounting fixes, and SQLite contention elimination (phases 1-5)

DarriEy:
## What this branch delivers

**Read-only web dashboard (phase 1 of the frontend roadmap)**
- FastAPI service (`auramaur/web/`, optional `[web]` extra) + React/Vite SPA (`web/`) + loopback-only `auramaur-web` compose service.
- Read-only by construction: SQLite `mode=ro` + `query_only`, no venue credentials in the container, own entrypoint that stays up through KILL_SWITCH.
- Paper/live book toggle (envelope serves BOTH books + the armed `bot_mode`), per-strategy realized P&L from `pnl_ledger`, category exposure, per-venue split, Kraken paper book panel.
- Bot-side venue balance recorder: the bot records venue cash to `venue_balances`; read-only consumers show it with an honest age.
- Never-a-dead-end UX: every failure mode (API down, schemaless DB, stale stream) explains itself and self-recovers.

**Kalshi accounting fixes (found by dogfooding the dashboard against the live DB)**
- `sync_positions` passed `limit=1000`; the SDK caps at 200 client-side — live position sync was fully dead (#314 regression).
- Settlements sweep fetched through a stale typed SDK model that drops the current API fields, and its P&L formula used the venue `revenue` field, which is always 0 on the current API (verified against live payloads) — every booking would have been a fabricated loss. Reworked: raw-JSON fetch; payout from our own cost_basis (token x size), venue single-sided cost as unambiguous fallback; two-sided histories with no record are reported, never guessed.
- Cockpit `_portfolio_pnl` joined cost_basis without `token`, fanning out two-sided positions and double-counting P&L (TUI included).

**SQLite contention elimination (docs/plans/db-contention-plan.md, all 5 phases)**
Root causes (verified live: ~300 "database is locked" errors/2 days across 13 subsystems): write transactions held across network awaits, multiple in-process writers, external openers running DDL against the live file, and no serialization on the shared connection (cross-task commit/rollback bleed).
1. Syncers/reconciler gather network data first, write once briefly; `synchronous=NORMAL`.
2. `Database.transaction()` — asyncio-locked `BEGIN IMMEDIATE`; adopted across non-money frequent writers; >250ms-hold warning.
3. LineageObserver shares the one writer connection (second connection deleted).
4. Read-only policy: agentmcp opens the bot DB transient `mode=ro`; all 25 CLI sites connect with `ensure_schema=False` (no DDL, no write locks); policy in CLAUDE.md.
5. `record_fill` atomic + retry-safe (dedupe inside the transaction); gateway commits scoped. **Soak-gated — see below.**

Also: call-budget counter no longer busy-waits on the event loop (31s stalls -> 250ms fail-fast with pending fold-in).

## Verification
- 1312 tests green on the merged branch (full-dependency environment).
- Phases 1-4 already deployed to the live bot via `deploy/db-phases-live` (this branch with phase 5 reverted): Kalshi sync green every cycle, zero locked errors since.
- Phase 5 is running a 48h paper soak in an isolated instance (own DB, paper gates, credentialed venues disabled). **Merge gate: this PR merges only after the soak passes** (0 partial fills, 0 money-path errors, 0 locked errors, transaction holds <1s).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,4 +14,6 @@ __pycache__
 .ruff_cache
 dist
 build
+web/node_modules
+web/dist
 deploy/ibgateway/ibgateway-*-standalone-linux-x64.sh
--- a/.env.example
+++ b/.env.example
@@ -77,12 +77,12 @@ AURAMAUR_LIVE=false
 # arm them explicitly on the final host only after verification.
 AURAMAUR_CONTAINER_LIVE=false
 AURAMAUR_CONTAINER_ENABLE_TRANSFERS=false
-# The dedicated IBKR live login is a read-only quote source. Port 4002 is an
-# intentional deployment-local choice; local paper execution remains separate.
-IBKR_QUOTE_ENVIRONMENT=live
+# IBKR is opt-in. Start with a paper Gateway/login; the socket stays read-only
+# and strategy fills remain in local paper ledgers.
+IBKR_QUOTE_ENVIRONMENT=paper
 IBKR_QUOTE_PORT=4002
-IBKR_ENABLED=true
-IBKR_MULTIASSET_PAPER_ENABLED=true
+IBKR_ENABLED=false
+IBKR_MULTIASSET_PAPER_ENABLED=false
 # 1=live. Multi-asset paper fills fail closed for 2/3/4.
 IBKR_MARKET_DATA_TYPE=1
 # Stable private addresses make the Gateway Trusted IP entry survive restarts.
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Shell scripts are executed inside Linux containers; a CRLF checkout on a
+# Windows dev machine (core.autocrlf=true) would break them at `#!/bin/sh`.
+*.sh text eol=lf
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,6 +19,7 @@ When making git commits, use `Assisted-by: Claude (Anthropic)` in the commit mes
 - Paper trading interception happens in `auramaur/exchange/paper.py`.
 - Risk manager in `auramaur/risk/manager.py` is the single gateway — no trade bypasses it.
 - The web dashboard (`auramaur/web/` + `web/` SPA) is read-only by construction: it opens the DB with SQLite `mode=ro` and must never gain venue credentials or order paths. Keep it that way.
+- Out-of-process DB consumers (web, MCP, scripts) open the trading DB via transient `mode=ro` URIs with `busy_timeout>=5000` and never run `Database.connect()`'s DDL against the live file; CLI tooling connects with `ensure_schema=False`.
 
 ## Code Style
 - Python 3.11+, async-first (asyncio)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,6 +18,7 @@ When making git commits, use `Assisted-by: Claude (Anthropic)` in the commit mes
 - All money flows through `auramaur/exchange/client.py` — this is the ONLY file that touches the CLOB API for orders.
 - Paper trading interception happens in `auramaur/exchange/paper.py`.
 - Risk manager in `auramaur/risk/manager.py` is the single gateway — no trade bypasses it.
+- The web dashboard (`auramaur/web/` + `web/` SPA) is read-only by construction: it opens the DB with SQLite `mode=ro` and must never gain venue credentials or order paths. Keep it that way.
 
 ## Code Style
 - Python 3.11+, async-first (asyncio)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,13 @@
 FROM node:22-bookworm-slim@sha256:6c74791e557ce11fc957704f6d4fe134a7bc8d6f5ca4403205b2966bd488f6b3 AS node_runtime
 
+# Web dashboard SPA — built here so the runtime image serves static files only.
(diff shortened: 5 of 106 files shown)


DarriEy (commented):
## Review — exceptional work, verified with receipts; merge after the soak AND one compose line

I verified the load-bearing claims empirically rather than trusting the (unusually good) writeup, and reviewed every money-path hunk plus a full audit of the web/CLI surface.

### Claims verified against the live system
- **Sync-dead (#314 regression): CONFIRMED.** The installed SDK declares `Field(le=200, strict=True)` on `limit`, and the live logs carry the exact repeating pydantic rejection of `input_value=1000` — position sync was dead from the #314 deploy until phases 1–4 went live at 03:22Z, after which `kalshi.positions_synced count=54` is green every cycle. For the record: the #314 review (mine) validated the cursor loop''s logic but never ran the limit against the installed SDK — this PR caught what I missed.
- **Settlements: CONFIRMED against a live payload.** A fresh authenticated fetch shows exactly what the module docstring claims: `revenue: 0`, `value: 0`, `yes_count_fp == no_count_fp` (pair count, not held side), per-side dollar costs present — and the very first record sampled is a *two-sided* history, the ambiguous case this rework correctly refuses to book. The #314 formula would have fabricated losses had the sweep ever booked; production was saved only by it never having run. The side-aware own-records-first / single-sided-venue-fallback / report-never-guess design is right, and the `_iso` normalization preserving `source_ref` idempotency across code versions is a detail most rewrites would have broken.
- **Contention: 51 lifetime "database is locked" errors in the current log; 2 in the 40 minutes since phases 1–4 deployed.** Not yet the claimed zero, but consistent with the un-adopted money writers being exactly phase 5''s territory.

### Money-path reads (all sound)
- `sync_positions` network-first/write-once refactor keeps the SQL identical while removing awaits from the write window.
- `Database.transaction()`: asyncio-locked `BEGIN IMMEDIATE`, bounded legacy-writer wait, >250ms hold warning, and an honestly documented `synchronous=NORMAL` tradeoff.
- Phase-5 `record_fill`: dedupe-inside-transaction closes the read-then-insert race, and removing the cross-task `rollback()` eliminates the discard-someone-else''s-writes hazard. Correctly quarantined behind the 48h soak — **the PR''s own merge gate stands: do not merge before the soak passes.**
- Cockpit token-join fix: straightforward and real.

### Security findings (web/CLI audit)
Confirmed: loopback-only publish, no `.env`/secret inheritance, GET-only API, `mode=ro` + `query_only`, whitelisted serialization, SPA same-origin only, agentmcp transient read-only. Two items need action:
1. **Pre-merge (one line): remove `auramaur-web` from the `broker` network.** It reads a file; it needs no network peers — and it currently sits adjacent to ibgateway, whose API auth is trusted-IP based. If the Gateway''s trusted list is ever the subnet instead of the bot IP, a compromised web container could place orders with no credential. This is the only order-capable vector in the container; deleting its `networks:` stanza (or an internal, gateway-less network) closes it.
2. **Document the rw-state-mount residual**: `./runtime/state` is mounted read-write for WAL/SHM, so `mode=ro`/`query_only` are SQLite-level, not kernel-level — a compromised web container could corrupt the DB or delete `/app/state/KILL_SWITCH`. Acceptable residual for a loopback phase-1, but it belongs in the doc (and the agentmcp docstring''s "kernel-enforced" wording overstates the same mechanism).

Smaller: `cli/manager.py` `_db()` still uses default `ensure_schema=True` (the "all 25 CLI sites" claim is off by one, and manager list-commands still take DDL locks); note that a version-skewed CLI (upgraded checkout, not-yet-restarted bot) will still migrate the live DB under the running bot — the no-DDL guarantee is steady-state only; the dashboard has no auth beyond loopback (fine for phase 1, worth one sentence in the doc).

### Verdict
The accounting fixes alone justify this PR — they''re verified-real regressions in money code, found by dogfooding, fixed fail-closed. Merge sequence: (1) soak passes per your own gate, (2) the one-line network removal, (3) rw-mount sentence in the doc. Items can land as a fast-follow if you prefer, but #1 is your own contract and #2 is one line — I''d do both before the button.

DarriEy:
All four review items applied in 4969998:

1. **Network isolation (pre-merge item)**: `auramaur-web` is off the `broker` network entirely — it lands on the compose default network purely for the loopback publish and can no longer reach ibgateway at all.
2. **rw-mount residual documented**: web/README.md now states plainly that `mode=ro`/`query_only` are SQLite-level (not kernel-level) guarantees given the rw state mount, and that loopback binding is the entirety of the dashboard''s auth. The agentmcp docstring''s "kernel-enforced" wording is corrected to match.
3. **`cli/manager.py` `_db()`** now passes `ensure_schema=False` — it was the one site the phase-4 sweep missed because it merged in from origin/main *after* the sweep ran (the "25 sites" count was true at sweep time).
4. **Steady-state caveat** added to the plan doc: a version-skewed CLI against a not-yet-restarted bot still runs migrations; restart the bot before CLI use after upgrading the checkout.

On the contention numbers: agreed the 2 residual locked errors in the 40-minute window are expected — the un-adopted money writers are precisely what the soaking Phase 5 wraps. The 7-day per-subsystem metric after full deployment is the real scoreboard.

The merge gate stands as written: the automated checker merges only after the 48h soak passes (2026-07-22 03:25 UTC), and `--auto` additionally waits for CI. Live redeploy remains human-triggered.

PR #329, opened 2026-07-21
feat(nlp): add local Ollama LLM tier

DarriEy:
## Summary

- add an optional local Ollama LLM tier for evidence distillation, news triage, embeddings, and ensemble analysis
- persist local-model telemetry and distilled claims with a v35 schema migration
- keep the tier evidence-side, shadow-safe, circuit-broken, and fail-open
- expose host Ollama to the container without moving configuration into environment variables
- make shared market-claim locking portable to Windows and ignore generated web artifacts

## Verification

- `python -m ruff check .`
- `uv lock --check`
- `git diff --check`
- `pytest -q`: 1,340 passed, 2 skipped
- focused local-tier tests: 34 passed, 1 skipped

PR #330, opened 2026-07-21
feat(research): add intelligence exploration evaluation

DarriEy:
## Summary
- add immutable timestamp-locked evaluation episodes, paired model/exploration runs, forecasts, outcomes, and proper scoring
- add deterministic single, multi-sample, and sample-plus-critic local Ollama treatments with independent seeds
- add execution-free virtual portfolios with bid/ask fills, fees, slippage, depth, limits, settlement, and risk metrics
- add a default-off shadow collector, venue-outcome settlement, schema v36 persistence, and read-only CLI scorecard

## Safety
- disabled by default
- no ExecutionGateway, RiskManager, exchange order, trades, or production portfolio path
- every treatment receives the same immutable episode payload
- historical/prospective cutoff is embedded in the frozen snapshot and prompt

## Verification
- Ruff: passed
- focused evaluation/local-LLM/settings suite: 56 passed
- full suite: 1395 passed, 3 skipped, 1 pre-existing failure
- the lone failure is tests/test_data_lineage.py::test_paper_cycle_lineage_through_resolution_and_report (cannot start a transaction within a transaction); reproduced unchanged on origin/main in a separate worktree
- git diff --check: passed

## Operation
Enable both local_llm.enabled and intelligence_eval.enabled. View resolved results with uramaur intelligence-eval.

### DIFF
--- a/auramaur/bot.py
+++ b/auramaur/bot.py
@@ -1627,6 +1627,11 @@ class AuramaurBot(
         if self.settings.agent_trader.enabled:
             tasks.append(asyncio.create_task(self._task_agent_trader(), name="agent_trader"))
 
+        if (self.settings.intelligence_eval.enabled
+                and self.settings.local_llm.enabled):
+            tasks.append(asyncio.create_task(
+                self._task_intelligence_eval(), name="intelligence_eval"))
+
         # Deadline-ladder curve reader (paper-forced; amortized reading edge)
         if self.settings.term_structure.enabled:
             tasks.append(asyncio.create_task(self._task_term_structure(), name="term_structure"))
--- a/auramaur/bot_strategy_tasks.py
+++ b/auramaur/bot_strategy_tasks.py
@@ -111,6 +111,25 @@ class StrategyTaskMixin:
                 log.error("agent_trader.cycle_error", error=str(e))
             await asyncio.sleep(interval)
 
+    async def _task_intelligence_eval(self) -> None:
+        """Prospective local-model evaluation; shadow-only and order-free."""
+        from auramaur.evaluation.service import IntelligenceEvalService
+        from auramaur.nlp import local_llm
+
+        service = IntelligenceEvalService(
+            self._components.db, self.settings, self._components.discovery,
+            local_llm.get_client(self.settings, self._components.db),
+        )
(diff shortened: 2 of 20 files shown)


DarriEy (commented):
Reviewed (ops session). This is well-constructed and safe to land:

**Checked and clean:**
- **Schema v35→v36**: correct numbering against post-#329 main; migration follows the additive TABLES-first pattern like v28/v34.
- **DB write discipline**: every store write is a short, db-only `transaction()` block (put_episode/put_run/put_forecast/put_outcome); network work (`_invoke`) is structurally separated from persistence in the runner — no awaits on Ollama inside a held transaction. This is exactly the contention-plan contract; no new bleed sources.
- **Execution isolation**: `evaluation/` has zero imports from exchange/gateway/risk paths (stated and verified); the virtual portfolio is genuinely execution-free. No route to a real order.
- **Gating**: task creation is double-gated (`intelligence_eval.enabled` AND `local_llm.enabled`), default-off in tracked defaults — fresh clones and CI can never run it.
- **CLI scorecard**: `ensure_schema=False` per the phase-4 CLI policy.
- The lineage test failure is pre-existing/order-dependent (independently confirmed in the ops session before this PR existed — passes in isolation, flakes in-suite). Not attributable here.

**Two non-blocking notes:**
1. The paired-treatment design (same immutable episode payload, independent seeds) is the right instrument for the graduation-ladder question the audit raised — consider eventually feeding the resolved eval outcomes into the same scorecard surface as `forecast_snapshots` so the two evidence streams are comparable.
2. `intelligence_eval_report` opens the live DB read-write (with `ensure_schema=False`). Policy-compliant as written, but the transient `mode=ro` URI pattern (like `agentmcp/market_data.py`) would be strictly safer for a pure reader.

Merge-ready once the safety gate reports.

DarriEy:
Addressed review note #2 in fea7527: uramaur intelligence-eval now opens the trading DB through ReadOnlyDatabase (mode=ro, query_only=ON, transient connection) instead of the normal Database path. Verified with Ruff and 19 evaluation/web read-only regressions. Note #1 (unifying this evidence stream with orecast_snapshots) remains a deliberate follow-up to keep this PR's evaluation ledger semantics isolated.

PR #331, opened 2026-07-21
feat(research): unify forecast evidence streams

DarriEy:
## Summary
- add canonical venue-qualified market outcomes and a v37 migration
- normalize raw, calibrated, intelligence-eval, and information-exploration forecasts through one semantic evidence view
- materialize versioned proper scores with fixed horizons and event-weighted summaries
- make ResolutionTracker the venue-truth writer while discovering unresolved evidence from every stream
- make forecast lineage persistence transactional so observers cannot see half-written evidence
- add an opt-in all-streams intelligence report

## Verification
- 104 focused tests passed
- 1401 passed, 3 skipped in the full suite
- Ruff passed
- git diff --check passed

## Stack
This PR is stacked on #330 (eat/intelligence-exploration-eval). Retarget it to main after #330 merges.

### DIFF
--- a/auramaur/cli/reporting.py
+++ b/auramaur/cli/reporting.py
@@ -16,7 +16,9 @@ from auramaur.cli._base import console, main
 
 
 @main.command("intelligence-eval")
-def intelligence_eval_report():
+@click.option("--all-streams", is_flag=True,
+              help="Compare production, intelligence, and information-trial streams.")
+def intelligence_eval_report(all_streams: bool):
     """Show the resolved, market-relative shadow evaluation scorecard."""
     async def _run():
         from auramaur.evaluation.store import EvaluationStore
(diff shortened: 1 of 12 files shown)


DarriEy (commented):
Reviewed (ops session). This is the right unification — it implements the evidence-stream convergence suggested in the #330 review, and it does it with the house discipline intact:

**Checked and clean:**
- **Base retargeted to main** (was still stacked on the merged #330 branch — done during review).
- **v36→v37**: correct numbering, standard additive pattern.
- **Write discipline**: OutcomeRepository.record and ForecastScoreMaterializer.refresh are both short transaction() blocks; evidence.py has no network imports, and the materializer refresh runs post-resolution with failures contained to a warning (a scoring failure cannot break settlement).
- **Ordering is right**: canonical venue truth records BEFORE the legacy calibration feed, so a calibration failure can no longer strand eval/lineage scoring; ResolutionTracker stays the single venue-truth writer.
- **record_prediction commit path** retired to transaction() — another legacy-commit site gone (that was on the follow-up list).
- **No interference** with the post-settlement persist_realized_pnl flow added 2026-07-20.

**Two non-blocking notes:**
1. The three new UNION arms in the unresolved-evidence discovery inherit the calibration arm's known weakness (audit 2026-07-20): rows for markets that will never resolve stay in the scan set forever, and each scan-set market costs a venue get_market call per 30-min cycle. Today the addition is negligible (8 eval markets, 0 snapshot markets), and outcome-recording is self-limiting — but the expiry/void path the audit recommended for calibration should cover these arms too when it lands.
2. forecast_snapshots.exchange defaults to 'polymarket' when empty in the event_key join — consistent with the price_history convention, fine — but worth keeping an eye on once Kalshi strategic forecasts accrue, since a mislabeled venue would silently mis-key outcomes.

Merge-ready once CI reports green.

PR #397, opened 2026-08-05
Make capital decisions benchmark-relative

DarriEy:
## Summary
- subtract cash-rate opportunity cost from prospective graduation P&L using committed stake and actual holding time
- rank allocation candidates by excess expected value through resolution
- refuse capital allocation when benchmark-relative expected value is non-positive
- keep the stronger behavior tracked-on while preserving permissive class defaults for isolated callers

## Verification
- python -m pytest tests/test_allocator.py tests/test_graduation.py tests/test_engine_cycle.py -q (27 passed)
- python -m pytest -q (2098 passed, 5 skipped)
- python -m ruff check ... on all touched Python files (clean)
- git diff --check (clean)

## Follow-up
The existing test-session aiosqlite leak remains visible and is intentionally scoped to PR 3 of the audit sequence.

### DIFF
--- a/auramaur/broker/allocator.py
+++ b/auramaur/broker/allocator.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from datetime import datetime, timezone
 
 import structlog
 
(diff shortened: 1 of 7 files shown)


DarriEy:
Re-review (Claude, 2026-08-05, after amendment a412c1a):

**Original review found two operator-level concerns; both are now resolved by measurement, not opinion.**

1. **Allocator entry veto**: replayed against the prior 14 days of live entries, the hurdle would have refused 7 of 13 — precisely the llm_kalshi long-dated cells whose worth the 2026-10-31 pre-registered cell review exists to adjudicate on evidence. The hurdle also prices hold-to-RESOLUTION cost the book demonstrably does not pay (2026-08-05: recent entries were exited within a day on repricing rather than held to resolution). Amendment gates the veto behind benchmark.allocator_cash_hurdle_enabled, tracked default OFF, dated comments naming the review that owns the flip. EV ranking and the ≤0 drop are dormant while disarmed; armed behavior unchanged and still tested.

2. **Graduation bar (require_cash_benchmark: true)**: kept armed as authored. Measured against the entire current settled evidence base: the total charge is negligible (13 rows, median hold 0.7 days). It only grows teeth when slow-resolving positions dominate evidence — exactly when it should. Schema guarantees resolved_at NOT NULL; zero nulls in the live join population.

Clock-safety re-verified: benchmark/graduation sections are outside every strategy_version hash; the three hashed risk keys untouched. Full suite on the amended branch: 2099 passed, ruff clean. Merge-base diff vs main contains only the intended package (no interference with the manual-trade sweep landed today).

**Verdict: MERGE-READY.** Merging on operator instruction.

PR #398, opened 2026-08-05
Align the documented minimum-edge contract

DarriEy:
## Summary
- document the effective tracked neutral minimum edge as 2.5% after fees
- explain that risk tolerance scales the gateway floor and strategies may impose higher bars
- preserve and document RiskConfig's conservative 5% no-YAML fallback
- add a regression test that keeps CLAUDE.md synchronized with loaded tracked settings

## Verification
- python -m pytest tests/test_settings.py tests/test_risk_manager.py tests/test_router_crossing.py -q (76 passed)
- python -m pytest -q (2095 passed, 5 skipped)
- python -m ruff check tests/test_settings.py (clean)
- git diff --check (clean)

## Scope
Documentation and contract test only; runtime thresholds are unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -51,7 +51,10 @@ When making git commits, use `Assisted-by: Claude (Anthropic)` in the commit mes
   two always applies — check the effective value, not this line.
 - Daily loss limit: $200
 - Max open positions: 500
-- Minimum edge: 5% after fees
+- Minimum edge (neutral tracked baseline): 2.5% after fees. The risk-tolerance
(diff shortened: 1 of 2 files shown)


DarriEy:
Review (Claude, session 2026-08-05): verified against tracked config — defaults.yaml min_edge_pct is 2.5, and live rejections cite 'minimum 2.50%'. The old CLAUDE.md line described RiskConfig's no-YAML class fallback, not the effective value. No value changes, outside every strategy_version hash input, and the contract test prevents future drift. Merging on operator instruction.

PR #399, opened 2026-08-05
Fail tests on leaked database connections

DarriEy:
## Summary
- make surviving open aiosqlite workers fail the pytest session after safe cleanup
- record the owning test and connection-creation stack for actionable diagnostics
- distinguish genuinely leaked handles from already-closed workers finishing an older aiosqlite polling tick
- close the dashboard read-only connection if cancellation lands during connection initialization
- preserve original missing/unreadable database errors when cleanup itself cannot run

## Root cause
The dashboard broker could be cancelled after aiosqlite started its worker but before the awaited handle was assigned to ReadOnlyDatabase._db. Shutdown then had no handle to close, leaving a non-daemon worker alive.

## Verification
- focused lifecycle/leak regressions: 5 passed
- python -m pytest -q: 2105 passed, 5 skipped, with no reaper output
- focused Ruff: clean
- git diff --check: clean

The existing FastAPI/httpx deprecation warning is unchanged and out of scope.

### DIFF
--- a/auramaur/web/db.py
+++ b/auramaur/web/db.py
@@ -10,6 +10,8 @@ needs, so ``monitoring.cockpit.gather_state`` runs unchanged on top of it.
 
 from __future__ import annotations
 
+import asyncio
+from contextlib import suppress
 from pathlib import Path
 
 import aiosqlite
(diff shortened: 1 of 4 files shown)


DarriEy:
Review (Claude, 2026-08-05): **the load-bearing claim verified empirically** — full suite run on this branch REBASED OVER CURRENT MAIN (today's five commits: settlement mappers, manual-trade sweep, #397, #398): 2111 passed, exit 0, ZERO reaper output. The dashboard connect-race fix (hold the handle before awaiting; close on cancellation/failure before re-raising) eliminates the last known leak source — very likely the 2-connection session-end reap observed on the phase-3-era tree. With that gone, escalating the be600db backstop from reap-and-warn to reap-and-FAIL is exactly right: a future leak turns CI red in 90 seconds with the owning test and construction stack named, instead of hanging 6 hours.

Implementation notes from line-by-line: CancelledError correctly listed beside Exception (BaseException since 3.8); the leak regression test correctly exempts itself by reaping in-test; the id()-keyed stack registry cannot mislabel live leaked connections (id unreusable while alive). One cosmetic nit, no action needed: connections created during fixture teardown attribute to the prior test's nodeid — the recorded stack still points at the true construction site.

Verdict: MERGE-READY.

PR #400, opened 2026-08-05
Block unapproved dependency audit findings

DarriEy:
## Summary
- replace advisory-only pip/npm audit steps with JSON report generation plus a blocking policy verdict
- require exact advisory identity, owner, reason, and future expiry for every exception
- fail on malformed reports/policies, unapproved findings, and expired exceptions
- keep npm's blocking threshold at high while treating all pip-audit findings as blocking
- document the exception workflow and cover both report formats/policy states
- upgrade brace-expansion 5.0.8 -> 5.0.9 to fix the current high-severity GHSA-rgw5-rvv9-x895 finding rather than excepting it

## Verification
- real pip-audit 2.10.1: no known vulnerabilities
- real npm audit after lock update: 0 vulnerabilities
- policy tests: 8 passed
- full Python suite: 2118 passed, 5 skipped
- web: 14 tests passed, build passed, ESLint passed
- Ruff, workflow YAML, policy JSON, and diff checks: clean

## Policy posture
The initial exception list is empty. New findings fail CI unless remediated or explicitly accepted with accountable, expiring metadata.

### DIFF
--- a/.github/dependency-audit-policy.json
+++ b/.github/dependency-audit-policy.json
@@ -1,0 +1,4 @@
+{
+  "version": 1,
+  "exceptions": []
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,14 +27,12 @@ jobs:
           uv export --frozen --extra kalshi --extra ibkr --extra web
           --no-hashes --output-file requirements-audit.txt
 
-      # ADVISORY, never blocking. This verdict depends on an external advisory
-      # feed, so a HIGH published against any locked package would otherwise
-      # turn every open PR red with no code change — the same failure mode as
-      # the unpinned ruff that broke every PR when 0.16.0 shipped. pip-audit
-      # itself is pinned for that same reason.
-      - name: Audit locked Python dependencies
-        continue-on-error: true
-        run: uvx pip-audit==2.10.1 -r requirements-audit.txt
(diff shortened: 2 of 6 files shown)


DarriEy:
Review (Claude, 2026-08-05): this deliberately reverses the 'ADVISORY, never blocking' decision I wrote into ci.yml — and it earns the reversal. That comment protected against the ruff-0.16 class of failure: tool-content changes reddening every PR with no recourse. A CVE against the locked runtime deps of a live-money trading bot is different in kind — it should be loud — and the exception policy converts 'red with no recourse' into 'red until a two-minute triage commit' with accountable, expiring metadata. The brace-expansion fix (rather than exception) is the right opening move and proves the advisory-only audit had been silently tolerating a real HIGH.

Verified: full suite on this branch rebased over current main (2119 passed, exit 0); ruff clean on the new script/tests; the PR's own audit job ran WITH the blocking gate active against an empty exception list and passed in 26s — the audited sets are clean today. Malformed-report handling errors loudly as documented.

One consciously-accepted availability trade-off worth naming: an advisory-feed/registry outage now fails the report-generation step ('test -s') and reddens CI on unrelated PRs. Rare, self-identifying by step name, and it fails toward caution — acceptable for this repo; a retry-once on generation would be a fine future hardening, not a blocker.

Verdict: MERGE-READY.

PR #401, opened 2026-08-05
Add coverage and generative risk gates

DarriEy:
Enforces the measured 66 percent repository coverage floor in CI. Adds Hypothesis properties and a rule-based allocation state machine. Also caps Kelly sizing at available bankroll after the generative suite exposed that missing defensive bound. Verified: 2115 passed, 5 skipped, 66.3 percent coverage; focused suite 11 passed; Ruff and diff checks clean.


DarriEy:
Review (Claude, 2026-08-05, incl. amendment d9b3b35):

**The PR's substance is good and verified**: the Kelly bankroll bound is exactly the kind of defect generative testing exists to surface — it never binds in production (the stake ceiling sits far below bankroll) but closes a real hole for any future call site; kelly.py is outside every strategy_version hash, so clock-safe. The property tests are well-shaped (finite/non-negative/capped invariants, YES/NO mirroring, and a stateful capital-conservation machine with bounded examples), and hypothesis lands dev-extra only — the runtime image is untouched.

**One merge-blocker found empirically and fixed on the branch**: the 66% floor was measured at the branch point (66.3%), but five merges later, today's main measures **65.86%** in the CI-identical uv --frozen environment — the ratchet was red on arrival. Amendment d9b3b35 calibrates the floor to 65 (still fails any real regression; headroom for cross-environment measurement drift) and rewrites both comments so future raises follow a fresh measure on main, never speculation. Re-verified after amendment: 2108 passed, coverage 65.86% >= 65.0%, exit 0.

Verdict: MERGE-READY as amended.

PR #402, opened 2026-08-05
Bound concurrent paper experiments

DarriEy:
Winds down settlement_arb and the Kalshi resolution_lens arm after their preregistered two-week clocks expired, while preserving historical evidence and normal position exits. Adds a fail-closed 12-family paper-experiment capacity limit; tracked defaults now use 11 slots. Documents start/review/wind-down requirements and adds regression tests. Verified: 2123 passed, 5 skipped; 35 focused settings tests passed; Ruff and diff checks clean.

### DIFF
--- a/config/defaults.yaml
+++ b/config/defaults.yaml
@@ -1147,8 +1147,12 @@ settlement_arb:
   # is ALREADY PUBLISHED and deterministically satisfies/fails the criterion, and
   # the market hasn't repriced (the lag is the edge). No forecasting; the LLM only
   # extracts the predicate (adversarially verified, cached). PAPER-FORCED, own
-  # graduation cell. DEFAULT OFF — flip enabled:true to start the 2-week spike.
-  enabled: true      # spike STARTED 2026-06-26 — 2wk paper measurement
+  # graduation cell. The spike ran 2026-06-26 through its preregistered
+  # 2-week endpoint (2026-07-10). WOUND DOWN 2026-08-05 pending adjudication:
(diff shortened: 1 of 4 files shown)


DarriEy:
Review (Claude, 2026-08-05): **governance-as-code done properly, MERGE-READY** — with two findings the PR body could not have seen, both verified against the operator config and live DB, neither blocking:

1. **Hidden clock cost, measured and acceptable**: resolution_lens.kalshi_enabled lives inside the hashed resolution_lens section (only IBKR sections get _scope_config'd), so this flip re-versions the STILL-RUNNING Polymarket lens family and restarts its 14-day holdout. Measured cost: the family's current version is 6 days old with THREE decision snapshots — the reset discards almost nothing. The wound-down Kalshi arm produced 2 decisions on its current version, which rather proves the capacity thesis. Worth knowing this class of edit carries that cost; here it is trivially affordable.

2. **Boot safety under the runtime override verified**: the fail-closed validator runs at Settings construction, and the live bot boots with the gitignored operator config, not tracked defaults. Probed with the real runtime file: 8/12 active paper trials (live-promoted strategies correctly drop out of the paper count) — four slots of headroom, next deploy boots clean.

Both wind-down clock claims verify from the diff's own removed comments (spike started 2026-06-26; Kalshi restart 07-03 → 07-17). Evidence retention claim is structurally sound — exits/settlements run engine-level, not pillar-level. Full gauntlet rebased over current main: 2114 passed, coverage 65.87% ≥ 65.0, ruff clean.

Verdict: MERGE-READY.

PR #406, opened 2026-08-05
Redact credentials from data-source error logs

DarriEy:
## The leak

`aiohttp`'s `ClientResponseError` stringifies to include the full request URL. Three sources in `official.py` pass their credential as the **first** query parameter and then log `str(exc)[:120]`:

| site | param | exposure |
|---|---|---|
| `:185` Congress | `api_key` | 40-char key at chars 71–110 — **fully inside the slice** |
| `:159` BEA | `UserID` | 36-char GUID — fully inside |
| `:224` EIA | `api_key` | ~30 of 40 chars survive |

Reproduced against the installed aiohttp 3.14.3:

```
LOGGED: 403, message='Forbidden', url='https://api.congress.gov/v3/bill?api_key=AbCd…ABCD&format=
KEY LEAKED IN FULL: True
```

The trigger is the *normal* failure mode for a metered API — 403 over-quota, 401 after rotation — and `logging.json_format: true` persists it to `auramaur.log`.

Truncation is not a mitigation: the key is early in the URL *because* it is the first parameter. `newsapi.py:71` escapes only by luck (`apiKey` is last and `[:80]` cuts before it). BLS is safe because its key is in the POST body.

## The fix

Fixing the three instances would leave the next source that passes a key as a query param to rediscover this. So: `redact_error()` in `data_sources/base.py`, applied to **all 27** exception-logging sites in the package.

- **Redaction precedes truncation**, so a short limit can't slice into a key the redactor would have covered.
- Known secret parameter names redacted by name (`api_key`, `apikey`, `key`, `token`, `access_token`, `auth`, `user_id`, `userid`, `secret`, `password`, `sig`, `signature`).
- A long-opaque-value backstop (`[A-Za-z0-9_-]{20,}`) covers parameter names nobody enumerated — verified by a test using a made-up `subscription_credential=`.
- Diagnostic value preserved: status, host and ordinary params (`format=json`, `limit=20`) still log.

## Tests

`tests/test_error_redaction.py` — 6 tests built on real `ClientResponseError` objects, not mocks: the Congress reproduction, BEA/EIA, the unknown-name backstop, redaction-before-truncation, non-destruction of ordinary params, and plain-string/non-HTTP inputs.

## Verification

- `tests/test_error_redaction.py`: 6 passed
- Full suite: **2112 passed, 5 skipped, 0 failed**
- `ruff==0.15.22 check` — clean

## ⚠️ Operator action required

This PR stops the leak going forward. It does not clean up what has already been written. **If Congress, BEA, or EIA have ever returned an error while those keys were configured, the keys are sitting in your log files in cleartext and should be rotated.** Worth grepping `auramaur.log*` for `api_key=` and `UserID=` to confirm scope.

### DIFF
--- a/auramaur/data_sources/aggregator.py
+++ b/auramaur/data_sources/aggregator.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING
 
 import structlog
 
-from auramaur.data_sources.base import DataSource, NewsItem
+from auramaur.data_sources.base import DataSource, NewsItem, redact_error
 if TYPE_CHECKING:
     from auramaur.lineage_observer import LineageObserver
 
@@ -112,9 +112,16 @@ class Aggregator:
                                    round((time.monotonic() - before) * 1000), "",
                                    datetime.now(tz=timezone.utc).isoformat(), mode))
                 return items
+            # 2026-08-06: these two rows are the package's catch-all sink. They
+            # catch whatever propagates out of *any* source — including a source
+            # that forgets to redact its own logging — and their 500-char window
+            # is four times wider than any per-source site. They also outlive the
+            # log: the text lands in source_fetches.error in the trading DB, which
+            # the read-only web dashboard renders. Redact before it is durable.
             except TimeoutError as exc:
                 fetch_rows.append((run_id, source_name, "timeout", 0,
-                                   round((time.monotonic() - before) * 1000), str(exc)[:500],
+                                   round((time.monotonic() - before) * 1000),
+                                   redact_error(exc, 500),
                                    datetime.now(tz=timezone.utc).isoformat(),
                                    getattr(source, "information_mode", "production")))
                 logger.warning("aggregator_source_timeout", source=source_name,
@@ -122,7 +129,8 @@ class Aggregator:
                 return []
             except Exception as exc:
                 fetch_rows.append((run_id, source_name, "error", 0,
-                                   round((time.monotonic() - before) * 1000), str(exc)[:500],
+                                   round((time.monotonic() - before) * 1000),
+                                   redact_error(exc, 500),
                                    datetime.now(tz=timezone.utc).isoformat(),
                                    getattr(source, "information_mode", "production")))
                 logger.exception(
--- a/auramaur/data_sources/base.py
+++ b/auramaur/data_sources/base.py
@@ -22,7 +22,11 @@ _SECRET_PARAM_RE = re.compile(
 )
 # Backstop for parameters we have not enumerated: any value that is long and
 # opaque is treated as a credential regardless of its name.
-_OPAQUE_VALUE_RE = re.compile(r"([?&][A-Za-z0-9_.-]+=)([A-Za-z0-9_-]{20,})")
+# 2026-08-06: '%' belongs in the value class. aiohttp percent-encodes the
+# reserved characters of a standard-base64 secret ('+' -> %2B, '=' -> %3D), and
+# without '%' the escape splits the value into runs shorter than the 20-char
+# threshold — so `?cred=AbCdEfGhIj%2FKlMnOpQrStUvWxYz0123` survived intact.
+_OPAQUE_VALUE_RE = re.compile(r"([?&][A-Za-z0-9_.-]+=)([A-Za-z0-9_%-]{20,})")
 
 
 def redact_error(exc: BaseException | str, limit: int = 120) -> str:
--- a/auramaur/data_sources/base.py
+++ b/auramaur/data_sources/base.py
@@ -2,10 +2,38 @@
 
 from __future__ import annotations
 
+import re
 from datetime import datetime, timezone
 from typing import Protocol, runtime_checkable
 
 from pydantic import BaseModel, Field, field_validator
+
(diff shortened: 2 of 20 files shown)


DarriEy:
Amendments applied and pushed as a fast-forward, 9e22713..67b485a (Claude, 2026-08-06) — no rebase, so the stacked #407 is unaffected. Container verification: 2113 passed, coverage 66% >= 65.0, ruff clean. Both new tests confirmed as genuine regression guards (stashing base.py alone fails exactly the percent test; stashing aggregator.py alone fails exactly the aggregator test).

1. **aggregator.py catch-all now redacted.** Both fetch_rows arms (timeout and generic Exception) route through redact_error(exc, 500). Persistence confirmed end to end: fetch_rows -> LineageObserver._write_ingestion -> INSERT INTO source_fetches(error) -> read back by monitoring/readiness.py. This is the sink that mattered most: catch-all for every source, 4x the window of any site the PR fixed, and it lands in the trading DB rather than a log file.
2. **Percent-encoding evasion closed** — value class now [A-Za-z0-9_%-]{20,}. Probes verified against installed aiohttp/yarl: a base64 key under an un-enumerated name went from fully visible to `cred=<redacted>`, while format=json / limit=20 / series=UNRATE survive byte for byte.
3. **rss.py:142 covered** (error=str(result) -> redact_error(result), 120 default matching the PR own bare-call sites).

**Corrections to my earlier review, for the record:**
- My stated mechanism was half wrong. yarl encodes + -> %2B and = -> %3D but leaves `/` RAW in a query value, so the %2F in my probe only occurs when a caller pre-encodes. The common params= path produces %2B/%3D. Adding % covers both — the fix stands, the reasoning needed fixing.
- **A related evasion remains open by deliberate choice:** `/` is still absent from the value class, so ?cred=<base64-with-slash> passes through unredacted. Adding `/` would redact ordinary path-shaped params (?redirect=/some/long/path) at real diagnostic cost, so it deserves its own decision rather than a silent widening. Recorded in the commit body as a known residual. Enumerated names (api_key, UserID, ...) are unaffected either way.
- **Still unredacted, out of scope:** logger.exception at aggregator.py:136 emits the full traceback whose last line carries the raw URL. Log file only, not the DB. Fixing it means restructuring exception logging.

Merging now, ahead of #407 which is stacked on this branch.

DarriEy:
## Follow-up: the opaque-value backstop was defeatable

While scoping the actual exposure, @DarriEy found that the backstop regex omits `/` — and it turns out to be worse than that. `[A-Za-z0-9_-]{20,}` is defeated by `/`, `+` **and** percent-encoding, because the run before the first such character falls under the 20-char threshold and nothing matches at all. Verified:

```
LEAKS  base64 with slash, unknown param name
LEAKS  base64 with plus
LEAKS  percent-encoded
ok     plain opaque (control)
```

All three leaked in full, not partially.

Fixed by terminating at a real delimiter — `[^&\s'"]{20,}` — the same terminator set the named-parameter rule already used. Four new parametrized tests cover base64-with-slash, base64-with-plus, percent-encoded and a dotted JWT, each under a parameter name deliberately not in the enumerated list, plus one asserting short diagnostic params still survive.

Worth recording why this happened: enumerating the characters a secret may contain is the same losing move as enumerating the ways untrusted text can be unsafe — which is the bug this same audit found in `sweep_invariants.py`'s `untrusted-repr` check, fixed the same way. Both enumerations were written from the author's imagination, and the gap was precisely what wasn't imagined.

## Scope correction on this PR

The audit's scan came back clean across log files (38 MB incl. 3 rotations), 18 error-bearing DB columns across 104 tables (2,572 rows carrying error text), and container stdout — **0 matches everywhere**. The reason is structural: Congress, BEA and EIA are not configured, so the leak path was never reachable. The only credentialed source in use is FRED, which goes through `fredapi`, whose exceptions don't embed the URL.

So this PR is prophylactic, not remediation. **No key rotation is required.** It closes the hole before those sources are ever enabled — including the aggregator sink that would have written the values into the trading DB.

DarriEy:
Reconciliation note — we fixed this concurrently.

67b485a landed while I was working on the same gap. I fetched, read it, and **rebased on top rather than force-pushing**; nothing of yours was discarded. Final order:

```
8d9ce0a  Stop enumerating the value class; terminate at the delimiter
67b485a  Close three gaps in data-source error redaction
9e22713  Redact credentials from data-source error logs
```

Two things your commit caught that mine did not, and they are the more important half:

- **`aggregator.py`** — the widest sink, and the only one that outlives the log file. Its error text is written to `source_fetches.error` and rendered by the read-only dashboard. I had scoped this to log files only and would have left the DB sink open.
- **`rss.py:142`** — a missed sibling in a file whose other site was already converted. Exactly the pattern this whole review kept finding.

My commit only supersedes the one line we both touched, and it answers the residual yours explicitly left open. Worth noting the failure mode was worse than `/`: `/`, `+` and `%` each caused the secret to leak **in full**, not partially — the run before the first out-of-class character is under the 20-char threshold, so the regex matches nothing at all rather than matching part of it.

Both test suites are kept intact — your aggregator-persistence test and the four parametrized backstop cases. 2119 passed, ruff clean.

PR #407, opened 2026-08-05
Make two safety gates fail closed instead of open

DarriEy:
Two guards that became **permissions** under the exact condition they exist to handle.

## 1. `risk/manager.py` — `force_paper` reported 2 of 6 restrictions

`is_paper_entry` is built from six sources and is what disables the live-only checks:

```python
is_paper_entry = (
    not self.settings.is_live
    or self.live_entries_blocked        # operational preflight BLOCK
    or self._paper_forced_strategy(signal.strategy_source)
    or cell.force_paper
    or force_paper                       # caller-declared
    or extreme_div.force_paper
)
```

The decision reported:

```python
force_paper=cell.force_paper or force_paper,   # two of six
```

`execution_gateway.py:112` computes `is_live = settings.is_live and not intent.force_paper`. So an entry could skip `check_category_allowlist`, `check_dispute_risk` and `check_divergence_band` **because** it was restricted, and still go to the venue with real money.

Concretely: preflight fails → `bot._run_live_gate` sets `live_entries_blocked` → the allowlist check is never appended → a non-allowlisted category is approved at full size, live. **Arming the guard made the system more permissive.**

The comment directly above that line already stated the correct invariant:

> Report the effective mode, not just the ladder's view — a caller-declared paper entry must come back marked paper or the pillar would submit it live after the gate judged it as paper.

It was implemented for two of the six inputs. There is now a single `paper_forced` expression used both to scope the checks and to populate the decision, so the two cannot drift again.

The global live gate is deliberately **excluded** from `paper_forced` — it is the gateway's own precondition, and folding it in would report "this entry was restricted" about a bot that simply isn't live. There's a test pinning that.

## 2. `exchange/kraken.py` — per-order USD cap skipped when the price probe failed

Safety layer #3 in that module's own docstring sat behind `if est_price:`:

```python
est_price = price or await self.get_price(pair)
if est_price:
    if notional > cap: return BLOCKED
```

`get_price` returns `None` whenever `_public("Ticker")` yields nothing, and `_public` **logs and returns `{}`** on any error rather than raising (`kraken.py:74-76`). A 5xx, a rate-limit response, or a Cloudflare blip removed the ceiling entirely and the order proceeded to `AddOrder` unbounded. Every autonomous caller passes `price=None`, so that probe is the sole source of the number the cap needs.

Now an unpriceable pair is refused — matching `transfers.py`, which fails closed at every equivalent point and is the model the rest of the codebase should follow here.

## Verification

Every new test was run against the **unfixed** code first:

```
--- against UNFIXED manager.py ---
FAILED test_preflight_block_is_reported_as_force_paper
FAILED test_extreme_divergence_is_reported_as_force_paper

--- against UNFIXED kraken.py ---
FAILED test_unpriceable_pair_is_refused_rather_than_uncapped
```

All pass after. Six tests added total, including one pinning that a paper-mode bot does *not* report `force_paper` (so the flag keeps meaning "restricted").

- Full suite: **2118 passed, 5 skipped, 0 failed**

### DIFF
--- a/auramaur/exchange/kraken.py
+++ b/auramaur/exchange/kraken.py
@@ -269,12 +269,23 @@ class KrakenSpotClient:
         #    any quote (USDC/USDT/EUR/…), not just USD-quoted pairs.
         cap = max_usd if max_usd is not None else self._settings.kraken.max_order_usd
         est_price = price or await self.get_price(pair)
-        if est_price:
-            notional = await self.usd_notional(pair, volume, est_price) or (volume * est_price)
-            if notional > cap:
-                return OrderResult(order_id="BLOCKED", market_id=pair, status="rejected",
-                                   is_paper=True,
-                                   error_message=f"order ${notional:.2f} exceeds cap ${cap:.2f}")
+        # Fail CLOSED when the pair cannot be priced. get_price returns None
+        # whenever _public("Ticker") yields nothing — a 5xx, a rate-limit
+        # response or a Cloudflare blip, since _public logs and returns {}
+        # rather than raising. Gating the cap check behind `if est_price:` meant
+        # a transient ticker failure removed the ceiling entirely and the order
+        # went to AddOrder unbounded. Every autonomous caller passes price=None,
+        # so the probe is the only source of the number this cap needs.
+        if not est_price:
+            return OrderResult(order_id="BLOCKED", market_id=pair, status="rejected",
+                               is_paper=True,
+                               error_message=f"cannot price {pair} to verify the "
+                                             f"${cap:.2f} per-order cap; refusing")
+        notional = await self.usd_notional(pair, volume, est_price) or (volume * est_price)
+        if notional > cap:
+            return OrderResult(order_id="BLOCKED", market_id=pair, status="rejected",
+                               is_paper=True,
+                               error_message=f"order ${notional:.2f} exceeds cap ${cap:.2f}")
 
         # 4. Three-gate live decision.  ``dry_run=False`` opens only the
         # per-order gate; it must never override either global live gate.
--- a/auramaur/risk/manager.py
+++ b/auramaur/risk/manager.py
@@ -213,14 +213,24 @@ class RiskManager:
                 strategy=signal.strategy_source, divergence=extreme_div.value,
                 threshold=rc.extreme_divergence_threshold,
                 model_prob=signal.claude_prob, market_prob=signal.market_prob)
-        is_paper_entry = (
-            not self.settings.is_live
-            or self.live_entries_blocked  # operational preflight BLOCK
+        # Every RESTRICTION that routes this entry to paper, as one value. This
+        # must be a single expression used both to scope the checks below and to
+        # populate RiskDecision.force_paper: when the two diverged, arming a
+        # protective latch (preflight BLOCK, extreme divergence) skipped the
+        # live-only checks while the entry still went out with real money —
+        # turning a guard into a permission.
+        paper_forced = (
+            self.live_entries_blocked  # operational preflight BLOCK
             or self._paper_forced_strategy(signal.strategy_source)
             or cell.force_paper
             or force_paper  # caller-declared, restriction-only
             or extreme_div.force_paper
         )
+        # The global gate is deliberately NOT part of `paper_forced`: it is the
+        # gateway's own precondition (is_live = settings.is_live and not
+        # intent.force_paper), so folding it in would say "this entry was
+        # restricted" about a bot that simply is not live.
+        is_paper_entry = not self.settings.is_live or paper_forced
 
         # Correlation, MODE-SCOPED. A paper entry adds NO real exposure, so it must
         # only correlate against the PAPER book — counting the live book would let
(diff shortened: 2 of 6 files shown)


DarriEy:
Review (Claude, 2026-08-06): **NEEDS-AMENDMENT (one line), plus a merge-order dependency.**

**Gate 1 (RiskDecision.force_paper) — unambiguously right, closes a live-money hole.** is_paper_entry scoped six restriction sources but the returned force_paper carried only two, and the gateway derives is_live = settings.is_live and not intent.force_paper. So an entry restricted by a preflight BLOCK latch, _paper_forced_strategy, or extreme divergence skipped the category-allowlist, dispute-risk and adverse-divergence checks and then went to the venue LIVE, AT FULL SIZE — arming a protective latch made the system strictly more permissive. Not really a fail-open conversion at all: a reporting-correctness fix making an existing latch mean its own name. No availability hazard — the failure mode is downgrade-to-paper, never rejection, the latch self-clears on the recheck timer, exits stay live throughout.

**Gate 2 (Kraken per-order USD cap) — right in principle, one residual.** The cap sat behind `if est_price:` while _public returns {} on error rather than raising, so any 5xx/rate-limit blip silently removed the ceiling and AddOrder went out unbounded. Refusing an unpriceable pair is correct. But this error class IS transient, and one caller can be hurt: the directional EXIT (treasury/kraken_pillar.py:826-830) passes no price=, firing a SECOND independent Ticker probe moments after the loop own probe at line 746 — if probe #1 succeeds and probe #2 blips, a stop-loss or trailing-stop exit is refused where it previously went through. Contained (position retained and retried next cycle, same shape as the existing exit_unconfirmed path; a sustained outage already skips the exit at 746), so amendment not block.

**Required amendment:** pass the already-validated price into the exit call — kraken_pillar.py:826-830, add price=price (the line-746 value already used for usd_notional on line 825). Deterministic cap on the number the loop already checked, redundant probe gone, fail-closed guarantee intact for callers that genuinely cannot price. Same optional treatment for the entry call at 954.

**Merge-order dependency:** this branch is STACKED ON #406 — it contains 9e22713 "Redact credentials from data-source error logs" (18 data_sources files + a 71-line test file) in addition to its own gate commit. The body describes only the gate work, so a reviewer reading it would not know ~60% of the changed files are another PR. #406 merges first; this branch then rebases (the gate commit touches manager.py/kraken.py only, so it should replay clean once the duplicate drops). Note your own review of the backstop regex matches mine — the [A-Za-z0-9_-]{20,} class misses percent-encoded base64 secrets; I am fixing that on #406 branch with two other gaps, so the amended version is the one to keep.

House rules clean: zero DB writes in the diff, no new transaction spans, triple gate and kill-switch ordering untouched. Clock-safety CLEAN — no defaults.yaml, no settings.py, no hashed section.

Evidence note for the ledger, not a blocker: entries that previously recorded live during preflight-block or extreme-divergence windows now record paper. Correct accounting, but it changes evidence composition mid-window for affected cells — same "never pool the eras" caveat as the 2026-07-29 llm containment fix.

Verdict: NEEDS-AMENDMENT — one line, then merge-ready behind #406.

DarriEy:
Amendment applied (956aa9c) and the branch rebased onto current main (Claude, 2026-08-06).

**Rebase:** #406 squash-merged as 3432f48, so this branch dropped its duplicated redaction commit via `rebase --onto main 9e22713`. It is now exactly its own two commits across six files — the gate work plus the exit fix — with no #406 content. Gauntlet on the rebased state: **2144 passed, coverage 65.94% >= 65.0, ruff clean.**

**The amendment**, and the agent applying it improved on my instruction in two ways worth recording:

1. **It declined to touch the entry call, correctly.** I had asked for the same price-pinning at ~954 since the precondition holds. It does — but on entry the cap is the BINDING ceiling and `vol` is derived from that same price (`size_for_usd(max_order_usd * 0.98, price)`), so `vol * price` is `0.98 x cap` by construction: passing the price would make the check compare a number against itself and turn the entry cap into a no-op except via the ordermin bump. A refused entry also costs nothing (next cycle retries, no exposure at risk), so the harm this amendment fixes is exit-specific. Reasoning is in the commit body; reversible with one kwarg.

2. **The same line closes a second latent bug neither of us listed.** The exit passes `max_usd = max(notional, max_order_usd) * 1.1` precisely so the cap never binds an appreciated winner — but `notional` came from the loop price while the check ran against a fresh probe, so a >10% upward tick between the two reads blocked the close outright. No transient failure required. Pinning the price closes that too.

Verification of the three preconditions I asked for: (a) both probes are `get_price(pair)` returning the side-agnostic last trade, no BUY/SELL asymmetry to lose; (b) the awaits in between are DB and signal calls that never mutate price, and `usd_notional` at 825 was already fed that same value, so the loop had committed to it before this change; (c) the refusal in kraken.py:279-283 is untouched and `price` is provably non-None and > 0 at the call site, so no None is smuggled past the gate — the entry, treasury and CLI callers still probe and still fail closed.

Regression proof: with the pillar change stashed, the two new exit tests fail with exactly the harm in the log — `reason=stop_loss status=rejected err='cannot price APEUSDC to verify the $27.50 per-order cap; refusing'`.

Two items flagged for separate looks, unchanged here: `_treasury`'s convert at kraken_pillar.py:455 has the same shape (validated price, not passed) but is a treasury order with no open exposure waiting on it; and `assess_kraken_spot` permits an ordermin bump to 1.5x max_order_usd while the entry passes no max_usd, so a bump between 1.0x and 1.5x is approved by the spec then rejected by the adapter.

Verdict: MERGE-READY. Merging now.

PR #408, opened 2026-08-05
Make the kill switch retire live exposure, and stop it swapping books

DarriEy:
Three separately-defensible decisions in three modules, composing into the opposite of what an operator pressing the emergency stop expects.

## What used to happen when you armed the kill switch

1. **Resting live orders kept filling.** `_check_kill_switch` only set `_running = False`. `_cancel_resting_live_orders` was reachable solely from `shutdown()` — which runs after `asyncio.gather(*tasks)` returns, and tasks that loop on `while True` rather than `while self._running` never return. GTC market-maker quotes and in-flight entries stayed working on the venue.
2. **Those fills were never recorded.** `_task_order_monitor` stops on the same `_running` flag.
3. **The cockpit silently switched to the paper book.** `gather_state` selected with `0 if settings.is_live else 1`, and `is_live` folds in `kill_switch_active`. Your live positions vanished from the screen.
4. **Nothing labelled the book**, so the swap was indistinguishable from "the positions are gone."

Net: the emergency stop left real orders working, stopped recording what they did, and showed you a different book.

**No individual decision here is wrong.** Folding the kill switch into `is_live` is *correct* for trading decisions. Reusing it for a *display* choice is the category error — and it's invisible from inside either module. `web/broker.py` got it right because it passes `book=` explicitly; `cli/run.py:88,107` don't.

## Fixes

**`bot.py`** — the cancel sweep now runs when the switch fires, not only at shutdown. Latched so it executes once across the ~20 `_check_kill_switch` call sites. A venue error during the sweep no longer blocks the halt (test pins this). `_cancel_resting_live_orders` now returns its count, surfaced in the critical alert so the operator knows what was retired.

**`monitoring/cockpit.py`** — book selection uses the *configured* live intent (`auramaur_live and execution.live`) instead of `is_live`. Which book the bot trades is configuration; whether it may trade *right now* is a separate question, and only the second is the kill switch's business.

**Header** — now names the book explicitly (`book: LIVE` / `book: paper`) alongside the existing mode indicator. `mode` answers "may I trade"; `book` answers "whose numbers are these". Conflating them is what made a halted-but-exposed bot read as flat.

## Tests

`tests/test_kill_switch_semantics.py` — 6 tests: book selection is invariant under the kill switch (parametrized both ways), a paper-configured bot still shows paper, the header names the book in the halted-live state, the sweep runs exactly once across repeated calls, and a venue failure during the sweep still leaves the bot halted.

## One test fixture changed

`test_cockpit_unified._fake_settings` was a `SimpleNamespace` without `auramaur_live` or `execution.live`. Real `Settings` always carries both — the fixture was under-specifying the object it stands in for. Added, with the paper-mode values that match its existing `is_live=False`.

## Verification

- Full suite: **2112 passed, 5 skipped, 0 failed**

### DIFF
--- a/auramaur/bot.py
+++ b/auramaur/bot.py
@@ -94,6 +94,9 @@ class AuramaurBot(
         # positions round-tripped from +$29/+$75 peaks to -$26/-$25.
         self._exit_failures: dict[str, float] = {}
         self._exit_pending: set[str] = set()  # Track exits with resting orders
+        # Latch so the kill switch's one-time cancel sweep runs once, not on
+        # every one of the ~20 _check_kill_switch call sites.
+        self._kill_switch_handled = False
         self._exit_gateway_obj = None  # Lazily built; see _exit_gateway property
         # Track attempted cross-exchange arbs so the scanner doesn't re-execute
         # the same opportunity every 5-minute cycle. Maps arb-key -> expiry ts.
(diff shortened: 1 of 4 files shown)


DarriEy:
Review (Claude, 2026-08-06): **the composite failure is real and this is the most important fix of the audit wave. MERGE-READY.**

All three factual claims verified against the code: (1) settings.is_live folds kill_switch_active (settings.py), so arming the switch silently swapped the cockpit to the paper book — and cli/run.py:88,107 call gather_state without book= while web/broker.py:85 passes it explicitly, exactly as the body says; (2) _check_kill_switch only set _running=False, and _cancel_resting_live_orders was reachable solely from shutdown(), which sits after a gather() that while-True tasks never let return — resting GTC exposure survived the halt, unrecorded; (3) the header had no book label, making the swap read as vanished positions.

The fix is correctly shaped: the latch is set BEFORE the awaited sweep, so concurrent callers cannot double-fire; a venue error logs and the halt proceeds (pinned by test); the alert now fires once with the cancelled count instead of spamming every check; the cockpit selector is precisely gates 1+2 of the triple gate — configured intent — leaving the switch to govern only permission. Paper-safety verified: the sweep only touches clients with a _live_pending ledger (hasattr guard) and cancels over the network BEFORE any DB write, no transaction held across venue calls.

One nit, not blocking: the two book-selection tests re-implement the selector expression rather than calling gather_state — the formula is tested, the wiring is covered indirectly via the fixed cockpit fixture. Fine as-is.

Full gauntlet rebased over current main: 2120 passed, coverage 65.91% >= 65.0, ruff clean.

Verdict: MERGE-READY.

PR #409, opened 2026-08-05
Size internal arb legs to a common share count, and label the NO leg

DarriEy:
`_execute_internal_arb` had three defects that made a "risk-free" package a guaranteed loss. It **auto-executes with no separate enable flag** (unlike `cross_exchange_auto_execute` / `negrisk_auto_execute`) and `arbitrage.enabled` defaults true.

## 1. Equal dollars, not equal shares

YES+NO of one market pays exactly **$1 per share** on the winning side, so the hedge exists only at equal share counts. Both legs were sized `position_size / price`:

| | yes 0.60 | no 0.35 |
|---|---|---|
| shares bought with $10 | 16.67 | 28.57 |

$20 out, worst case `16.67 × $1` back → **guaranteed −$3.33**, logged and alerted as an executed arb. Profit existed only when *both* legs priced under 0.50.

`_execute_negrisk_arb`, 150 lines above, already states the rule:

> Size all legs to a common **SHARE** quantity so each losing leg pays an identical $1.

Now applied here: `qty = min(position_size / yes_px, position_size / no_px)`, which also guarantees neither leg exceeds the risk-approved stake.

## 2. The NO leg had no `token`

`Order.token` defaults to `TokenType.YES` (`exchange/models.py:162`).

- **Kalshi** derives both the book side and the price inversion from `order.token`, *not* `token_id` — so this posted **two YES bids** instead of a hedge. The second usually tripped Kalshi's own resting-order dup guard and returned `SKIP_DUP`, leaving a single unhedged directional position reported as an executed arb.
- **Polymarket** routes purely on `token_id`, so the venue trade was correct — but `_record_result` writes `token=order.token` into `cost_basis`/`fills`. A NO holding landed under `'YES'`, which is the exact key `_exceeds_market_cap` reads, so cap accounting and marks were attributed to the opposite side.

## 3. Success reported unconditionally

`place_legs` runs these sequentially (`concurrent=False`) and either leg can independently return `INSUFFICIENT_BALANCE`, `POST_ONLY_REJECTED` or `SKIP_DUP`. The code logged `internal_executed` and sent "Internal arb executed" without inspecting either status — a half-leg read as success. Both sibling paths already raise `critical` on a partial fill; this one now does too.

## Deliberately NOT fixed here

The same equal-dollar sizing exists at `entailment_arb:602` and `cross_venue_arb:334`. Those route through `submit_paired(size_dollars=...)`, and the gateway picks each leg's execution price inside `router.route` / `prepare_executable_order` — Kalshi crosses the spread with an aggression term, Polymarket applies a min-order bump. **A pillar cannot predict the share count its dollars will buy**, so computing dollars from a quoted price would substitute one wrong answer for another.

Fixing those properly means giving `submit_paired` a common-share-quantity contract — a gateway *interface* decision rather than a pillar edit, and yours to make. Both consumers are paper/disabled today (`entailment_arb.paper: true`, `cross_venue_arb.enabled: false`), so this is latent.

## Tests

`tests/test_internal_arb_hedge.py` — 7 tests, including one that pins the *old* behaviour as a guaranteed loss on the reported case, a parametrized sweep showing equal-share sizing is never negative, a check that neither leg exceeds the approved stake, and a source assertion that the NO leg carries `TokenType.NO`.

- Full suite: **2113 passed, 5 skipped, 0 failed**

### DIFF
--- a/auramaur/bot_arb_execute.py
+++ b/auramaur/bot_arb_execute.py
@@ -268,30 +268,51 @@ class ArbTradeExecutionMixin:
 
         exchange_client = engine.exchange
 
+        yes_px = market.outcome_yes_price
+        no_px = market.outcome_no_price
+        if yes_px <= 0 or no_px <= 0:
+            return
+
+        # Size BOTH legs to a common SHARE quantity, matching
+        # _execute_negrisk_arb above. YES+NO of one market pays exactly $1 per
+        # SHARE held on the winning side, so the hedge exists only at equal
+        # share counts. Sizing each leg to the same DOLLARS bought unequal
+        # shares: at yes=0.60 / no=0.35 with $10 a leg that is 16.67 vs 28.57
+        # shares — $20 out, worst case 16.67 x $1 back, a guaranteed -$3.33 on
+        # a trade booked and alerted as risk-free. Profit existed only when
+        # BOTH legs priced under 0.50; above that the "arb" was a losing bet.
+        qty = min(position_size / yes_px, position_size / no_px)
+        if qty < 1:
+            return
+
         # YES leg
         yes_order = Order(
             market_id=market.id,
             exchange=exchange_name,
             token_id=market.clob_token_yes or market.id,
             side=OrderSide.BUY,
-            size=position_size / market.outcome_yes_price if market.outcome_yes_price > 0 else 0,
-            price=market.outcome_yes_price,
+            token=TokenType.YES,
+            size=qty,
+            price=yes_px,
             dry_run=not self.settings.is_live,
         )
 
-        # NO leg
+        # NO leg. token= is REQUIRED, not decorative: Order.token defaults to
+        # YES, Kalshi derives both the book side and the price inversion from
+        # order.token (not token_id), so omitting it posted two YES bids
+        # instead of a hedge — and the gateway writes token=order.token into
+        # cost_basis/fills, so a NO holding was booked under 'YES' and its cap
+        # accounting attributed to the opposite side.
         no_order = Order(
             market_id=market.id,
             exchange=exchange_name,
             token_id=market.clob_token_no or market.id,
             side=OrderSide.BUY,
-            size=position_size / market.outcome_no_price if market.outcome_no_price > 0 else 0,
-            price=market.outcome_no_price,
+            token=TokenType.NO,
+            size=qty,
+            price=no_px,
             dry_run=not self.settings.is_live,
         )
-
-        if yes_order.size < 1 or no_order.size < 1:
-            return
 
         try:
             (yes_result, _), (no_result, _) = await self._exit_gateway.place_legs(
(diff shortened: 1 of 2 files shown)


DarriEy:
Review (Claude, 2026-08-06): **MERGE-READY.** All three claimed defects verified real against main, and the first one is serious.

**Equal dollars, not equal shares** (bot_arb_execute.py:277,288): both legs sized position_size/price, so shares diverge whenever prices differ — the executor contradicted its own scanner, which emits at YES+NO < 0.97 and computes expected_profit_pct on the equal-share payout model. At the PR example ($10/leg, YES 0.60 / NO 0.35): 16.67 YES vs 28.57 NO shares, $20 out; YES resolves to $16.67 — **P&L -$3.33**. Precisely: worst case is negative whenever max(yes,no) > 0.50, i.e. nearly every real opportunity. So a "risk-free arb" alert was fronting a coin flip. Reachability is worse than a normal lane: internal arbs auto-execute with NO enable flag, unlike cross_exchange and negrisk.

**NO leg unlabeled** — verified end-to-end and this is the same defect FAMILY as yesterday double-booking. Order.token defaults YES; Kalshi derives book side and price inversion from order.token alone, so the old NO leg posted as a second YES bid at the NO price (usually eaten by the dup guard). On Polymarket the venue routing was correct but the gateway writes Fill(token=order.token) into fills/cost_basis, collapsing both legs into ONE blended row under YES — while the post-3f70692 reconciler writes the NO asset under NO. That bot-row/venue-row label disagreement is exactly the one-settlement-two-source_refs signature that produced the double-booking incident. 3f70692 fixed the venue-side normalizer; this fixes the bot-side generator. Complementary, not duplicative. Bonus the body omits: place_legs batch reservation keyed both legs under YES, summing them against the abs ceiling — the fix also stops that false block.

**Unconditional success report** — confirmed; the ok tuple matches the gateway _OK_STATUSES and every failure sentinel carries status rejected.

Fix correctness: qty = position_size/max(px) gives guaranteed profit qty*(1-yes-no) >= 0.03*qty at the trigger, each leg notional <= approved stake. No floor asymmetry (same fractional qty both legs; Kalshi rounds both identically; Polymarket submits size unmodified on this path). Execution contract intact — one place_legs call, registry-pinned. Clock-safe: code-only.

Follow-ups, none blocking: (1) the sizing tests exercise a local _qty reimplementation, so a production regression stays green — extend the existing mock harness in test_arbitrage.py to assert legs[0].size == legs[1].size and the tokens on real orders; (2) PR body says cross_venue_arb.enabled: false — tracked defaults.yaml:1355 has enabled: true, paper: true; the safety conclusion holds via paper, not disablement. Related and worth recording: cross_venue paper evidence is accruing under the still-unfixed equal-dollar submit_paired sizing, and since strategy_version hashes config not code, a later gateway fix will NOT reset its clock — do not pool evidence across that fix (the llm-evidence-epoch lesson); (3) the 0/2-legs case raises a critical alert claiming an unhedged position when nothing filled.

Verdict: MERGE-READY.

PR #411, opened 2026-08-06
Add prospective maker observatory

DarriEy:
## Summary
- add a shadow-only maker observatory for L1 microprice skew, top-five depth imbalance, decayed signed flow, volatility, and quote persistence
- persist versioned observation/fill provenance with restart-safe 30/60/300-second markouts, bounded lateness, retention, and schema migration v49
- enforce prospective warmup/holdout evaluation, credible-fill rules, clustered confidence intervals, explicit promotion blockers, and read-only CLI reporting
- keep every observatory read/write isolated from quote formation and execution authority

## Evidence contract
- synthetic resting paper fills are diagnostic only and never count toward promotion
- late marks remain visible but invalid
- report windows use fill time and distinguish pending, due, valid, and credible holdout marks
- any future quote-policy change requires a separately versioned experiment

## Verification
- full suite: 2132 passed, 5 skipped
- focused observatory/governance suite: 40 passed
- Ruff: clean
- git diff --check: clean

### DIFF
--- a/auramaur/bot.py
+++ b/auramaur/bot.py
@@ -1279,6 +1279,43 @@ class AuramaurBot(
                 show_error(f"Market maker cycle failed: {e}")
             await asyncio.sleep(interval)
 
+    async def _task_maker_observatory(self) -> None:
+        """Resolve due maker fill markouts on their own timer.
+
+        Deliberately NOT folded into `_task_market_maker`. The scan grows with
+        retained history and used to run inside `_quote_market`, inside the
+        per-market `op_timeout` window, holding the process-wide Database
+        serializer — 4.6 s per 5-market cycle at 45-day retention. Slow quotes
+        get picked off, so leaving it there meant the instrument would cause
+        the adverse selection it exists to measure. Its own task also keeps
+        resolution running when the maker has nothing to quote, so the
+        measurement record does not silently develop holes.
+
+        This task deliberately does NOT `beat()`. `check_strategy_data_delivery`
+        fails a readiness criterion for any heartbeat whose strategy has no
+        registered data contract, and this task consumes no market data — it
+        reads rows the maker already wrote. Its liveness surface is the
+        observatory's own valid-mark completeness, which is the honest one.
+
+        No kill-switch gate, on purpose: this places no orders and moves no
+        cash. Halting it would only corrupt the measurement record (marks would
+        go permanently late) for a reason unrelated to trading safety.
+        """
+        mm: MarketMaker | None = self._components.market_maker
+        observatory = getattr(self.settings, "maker_observatory", None)
+        if mm is None or observatory is None or not observatory.enabled:
+            return
+        interval = observatory.resolve_interval_seconds
+        while self._running:
+            try:
+                resolved = await mm.resolve_markouts()
+                if resolved:
+                    log.info("maker_observatory.markouts_resolved", marks=resolved)
+            except Exception as e:
+                # Measurement must never take the process down.
+                log.warning("maker_observatory.resolve_error", error=str(e))
+            await asyncio.sleep(interval)
+
     async def _task_price_monitor(self) -> None:
         """Monitor real-time price changes via WebSocket."""
         ws = self._components.websocket
@@ -1936,6 +1973,12 @@ class AuramaurBot(
         # Market maker (if enabled)
         if self._components.market_maker:
             tasks.append(asyncio.create_task(self._task_market_maker(), name="market_maker"))
+            # Shadow measurement, on its own timer so markout resolution never
+            # sits on the quoting path.
+            if getattr(self.settings, "maker_observatory", None) is not None \
+                    and self.settings.maker_observatory.enabled:
+                tasks.append(asyncio.create_task(
+                    self._task_maker_observatory(), name="maker_observatory"))
 
         # Hybrid strategy report (hourly per-pillar P&L)
         if self._hybrid and self._components.attributor:
--- a/auramaur/cli/reporting.py
+++ b/auramaur/cli/reporting.py
@@ -61,21 +61,40 @@ def maker_observatory_report(days: int):
                       "holdout marks. Synthetic paper fills never count.[/]")
         scorecard = Table(title="Frozen-threshold holdout scorecard")
         for column in ("Horizon", "Feature", "Warmup", "Holdout", "Markets",
-                       "High-low markout"):
+                       "Measured", "High-low markout"):
             scorecard.add_column(column)
         for row in features:
+            # "Measured" is the share of this horizon's marks for which the
+            # feature was recorded at all. A blank effect over low coverage is
+            # an absent result, not a negative one.
+            covered = row["coverage"]
             scorecard.add_row(
                 f"{row['horizon_seconds']}s", row["feature"],
                 str(row["warmup_n"]), str(row["holdout_n"]), str(row["markets"]),
+                ("—" if covered is None
+                 else f"{row['covered_n']}/{row['marks_n']} ({covered:.0%})"),
                 "—" if row["effect"] is None else f"{row['effect']:+.4f}",
             )
         console.print(scorecard)
         sampled = coverage["sampled_quote_coverage"]
+        flow = coverage["flow_coverage"]
         console.print(
             f"[dim]Cadence samples={coverage['samples']}; active quote present="
             f"{'—' if sampled is None else f'{sampled:.1%}'}; "
             f"holdout samples={coverage['holdout_samples']}. This is sampled "
             "coverage, not venue-certified uptime.[/]")
+        console.print(
+            f"[dim]Trade-feed coverage: {coverage['flow_samples']}/"
+            f"{coverage['samples']} observations had signed_flow data"
+            f"{'' if flow is None else f' ({flow:.1%})'}. The rest are NULL — "
+            "no feed reached the market, which is NOT balanced flow and is "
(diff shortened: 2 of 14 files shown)


DarriEy:
Heads-up before this leaves draft (Claude, 2026-08-06): **schema version collision.** main is at SCHEMA_VERSION 48. This PR implements `_migrate_v48_to_v49`, and so does PR #417 (deterministic arb pair post-check), which is reviewed and merge-ready. If both land as written, the class ends up with two methods of the same name — Python keeps the last definition and one migration silently never runs, with no error at any layer. #417 keeps v49; **please renumber this one to v50** (models.py SCHEMA_VERSION, the migration method name, and its call in the chain).

Separately, from a first pass over the diff, one item to fix while you are in there: the nine `observatory_*` keys land INSIDE the `market_maker:` section of defaults.yaml, which is hashed into that strategy's `strategy_version`. A PR whose stated contract is "no feature below changes quotes" would therefore re-version the strategy. It costs nothing today — market_maker records zero decision_snapshots, so there is no clock to reset — but that fact is worth knowing on its own (the maker cannot currently graduate, because it produces no prospective evidence). Moving the keys to their own `maker_observatory:` section keeps the shadow-only claim literally true.

Full review to follow once it is out of draft.

DarriEy:
Review + fixes applied (d995c48), rebased onto current main, still draft (Claude, 2026-08-06). **2291 passed, coverage 67.16% >= 65.0, ruff clean.** The metrics are correct and the restart-safety claim holds in code — but two findings change what this PR is, and one of them is the headline.

**CRITICAL, fixed: the migration collision was real.** `_migrate_v48_to_v49` was defined **twice** in `Database` once #417 took v49 on main. Python keeps the last definition, so **#417's arb post-check migration would have silently never run** — a data-layer failure with no error at any layer. Renumbered to v50 and verified from a real v48 database that it reaches 50 with both #417's postcheck columns AND the observatory tables, no duplicate method names, idempotent on re-run.

**HIGH: "no feature below changes quotes" is false as a COST claim.** `observe()` runs inline in `_quote_market`, inside the `op_timeout` window, and every statement takes the process-wide `Database._txn_lock` shared by ~30 pillar tasks. Measured on idle NVMe at the tracked config (max_markets=5, refresh 30s, paper => 2 instant fills/market/cycle):

| retained fills/market | observe() | 5-market cycle |
|---|---|---|
| 0 (day 0) | 7.6 ms | 38 ms |
| 40k (day 7) | 236 ms | **1.2 s** |
| 121k (day 21) | 452 ms | **2.3 s** |
| 259k (day 45) | 917 ms | **4.6 s** |

93% of that is `_mark_due`, whose predicate `unixepoch(?) - unixepoch(f.filled_at) >= ?` is non-sargable — EXPLAIN confirms the index serves `market_id=?` only, so every refresh rescans every retained fill for that market. **Book depth is not the problem** (compute_maker_features is 0.14 ms); volume is: ~1.3M fill rows + ~3.9M markout rows at 45 days, roughly **590 MB added to a live DB already at 1382 MB**. An observatory that makes a quote arrive four seconds late has changed the quotes.

**HIGH, fixed: `prune()` could never run.** The cycle counter was in-memory and reset to 0 on every start, firing only at cycle 2880 (24.0 h). Any restart tighter than that means retention never runs and the scan above grows without bound — and this is not hypothetical: the bot has RestartCount=1 and restarted again today at 12:17. Re-keyed to wall clock, prunes on the first cycle after startup.

**MEDIUM, fixed:** the nine `observatory_*` keys sat in `market_maker:`, which `_capture_decision` hashes, so the PR re-versioned the strategy it claims not to touch. Moved to a top-level `maker_observatory:` section, proven clock-neutral by hash rather than inspection: main `d4f1270c...`, PR as authored `c8ef0ba3...` (re-versioned), after fix `d4f1270c...` — identical to main.

**MEDIUM, fixed:** `MakerObservatory.record_fill` matched `exposure_registry.SENSITIVE_METHODS` by attribute name, forcing a pure measurement write to register as a `prediction_quoting` exposure-mutation callsite. Renamed `record_observed_fill`.

**Metrics verified correct** — microprice is the standard size-weighted form, depth-imbalance sign agrees with it, decay units are right (exactly 0.5 at one half-life), and the markout signs are correct on **both** legs including the ask leg's `1-p_NO` conversion. Late marks genuinely record `is_valid=0`. The evidence contract holds: `_prospective_stats` reads only decision_snapshots and never joins these tables. The CLI is genuinely read-only (mode=ro + query_only=ON, verified by running it and asserting zero writes).

**Mutation testing — 9 of 20 mutations survived the PR's own tests.** The worst is M4: adding `"synthetic"` to `CREDIBLE_FILL_EVIDENCE` — i.e. making paper fills promotable, precisely what the evidence contract forbids — left all six tests green. The `credible_holdout_marks == 0` assertion passed because its fixture was pre-holdout, not because of the evidence filter. Also surviving: paper fills in the feature report, blockers counting paper as live, completeness hardcoded to 100%, the holdout filter, the ask-leg conversion, `observe()` never being called from the quoting loop at all, `prune()` as a no-op, and a fabricated confidence interval. Tests 6 -> 20, all 20 mutations now killed.

**For operator judgment — deliberately not touched:**
1. **The O(N) growth is a design question, not a bug fix.** A sargable rewrite buys only 22%; the real options are a per-horizon watermark, shorter retention, or not persisting every synthetic fill — all change what is stored or measured.
2. **`signed_flow` may be structurally dead.** `OrderFlowTracker` is fed only by the websocket `on_trade` for the first 20 discovered markets, capped at 50 trades each; the MM's five markets are chosen by spread and usually are not in that set. The column can be identically 0.0 for reasons unrelated to flow, and nothing distinguishes "balanced" from "no feed" — the same failure mode as the qwen3 null-result, where a metric was structurally incapable of moving. Fixing it needs a nullable column.
3. **`_PAPER_TRIAL_SECTIONS` still omits market_maker** despite it being enabled+paper. Adding it could push active trials past the cap and raise at `Settings()` load — i.e. the bot would not start.
4. Fills with NULL observation_id are dropped by the summary's inner join — invisible rather than counted as incomplete.

DarriEy:
Amendments applied (d43c6db), still draft. **2305 passed, coverage 67.32%, ruff clean.**

**Markout resolution is off the quote path, and the numbers hold flat.** Same harness as the earlier measurement:

| retained fills/market | before | after | offline resolver pass |
|---|---|---|---|
| 0 (day 0) | 21.4 ms | 13.1 ms | 20.6 ms |
| 40k (day 7) | 744.6 ms | 13.9 ms | 24.0 ms |
| 121k (day 21) | 2.21 s | **13.6 ms** | 21.2 ms |
| 259k (day 45) | 4.49 s | **13.1 ms** | 27.3 ms |

Flat at 13-14 ms at every volume, and `observe()` drops from 6 serialized round-trips to 3. It lives in its own bot task on a 60 s timer rather than inside `check_fills` — correctly rejected folding it into the maker loop, since that runs under `op_timeout` and a multi-second scan would have eaten the fill-poll budget. Its own task also keeps resolution running when the maker has nothing to quote, so the record cannot silently develop holes.

**Semantics preserved by construction, not by cadence.** A mark is taken from the first observation at or after `filled_at + horizon` — exactly the book the inline scan used, since that scan ran on the observation that first found the fill due. The mark is therefore a pure function of the record, so resolver timing cannot change what it concludes. Horizons, lateness, validity and `is_valid=0` on late marks are untouched.

**Three corrections to my brief, all of which improve on it:**
1. **Sargability alone would NOT have fixed this.** In steady state nearly every retained fill satisfies any `filled_at` bound, so the selective predicate had to be "still owes a mark" — hence `marks_pending` plus a partial index, not just a rewritten WHERE. My brief asked for the weaker fix.
2. The old plan was not a full table scan — it seeked `market_id` then walked every fill in that market with a per-row markout probe. Same cost, different mechanism; wording corrected in code, tests and docs.
3. Day 0 after the change is ~13 ms, not the ~38 ms I quoted as a floor: that figure already included the inline scan, which cost round-trips even against an empty table.

**A hazard it introduced and then caught itself:** the batch ceiling drains oldest-first, so fills stranded by a market leaving the maker's five would have held the head of the queue until retention pruned them 45 days later, starving every newer mark. The scan now skips fills whose market has not been observed as far as their earliest horizon; they stay pending and still take their late/invalid mark if the market returns.

**`signed_flow` is now honest.** `OrderFlowTracker.signed_flow()` returns None (never 0.0) when no trade was ever seen for the market — a market the feed HAS reached that nets to zero still scores 0.0, which is a measurement. Column nullable, and every consumer updated: the feature report emits `marks_n`/`covered_n`/`coverage` per (horizon, feature), quote coverage adds `flow_samples`/`flow_coverage`, and the CLI shows a Measured column plus a yellow warning whenever coverage < 100% explaining the 20-market subscription limit. That closes the qwen3-null-result failure mode — absence can no longer read as balance.

**27 mutations, 27 killed**, including inline resolution restored, marks fabricated for unmarkable fills, `marks_pending` cleared while horizons remain, no-feed coerced to 0.0, coverage counting NULLs, and the starvation guard removed. Notably the agent distinguished a first-draft mutation that was a genuine no-op from a surviving one, and added a test specifically because two mutations initially had nothing that could catch them.

**Remaining for operator judgment** (unchanged, deliberately untouched): market_maker still absent from `_PAPER_TRIAL_SECTIONS`; fills with NULL observation_id are dropped by the summary inner join; and the v50 number now collides with #419 — whichever merges second must renumber, and #419 is currently blocked behind a separate exit-path fix.

PR #412, opened 2026-08-06
Stop booking resting and refused orders as positions

DarriEy:
Three shapes of one defect, all feeding the ledger the **graduation ladder** reads to authorize live capital: a resting order booked as a position, a refused order stamped as a fill, and — the mirror image the first fix opens — a real fill that books no position at all.

## 1. `exchange/paper.py:215` — a refused order reported as filled

```python
result = await self.execute(order, force=True)
result.order_id = order_id
result.status = "filled"        # unconditional
```

`force` bypasses only the marketability/resting check — **not the balance gate below it**. A trade-through on a book without enough paper cash returns `status="rejected", filled_size=0, filled_price=0` and updates no position, balance or cost basis. That object was then stamped `"filled"`.

`bot_order_monitor` booked it: a `fills` row at size 0 / price 0, plus `mark_fill(evidence="trade_through")`. `"trade_through"` is in `graduation.credible_fill_evidence`, so `_prospective_stats` counted the decision as an **executable holdout observation** — precisely what `require_executable_fills` exists to prevent.

**Severity, measured (2026-08-06, live DB):** this defect has **zero historical instances**. No `fills` row has `size <= 0` or `price <= 0` (0 of 0), and none of the 61 `trade_through` decision snapshots has `filled_price <= 0`. The path was reachable but never taken — a latent hole, not a corrupted record. The order *did* vanish, though: `continue` skipped the re-queue, so it neither filled nor rested.

## 2. Nine pillar call sites wrote a full-size record on `status="pending"`

Live Polymarket `place_order` **always** returns `pending`/`filled_size=0`, and maker-priced paper orders are always deferred. "pending" is the normal case, not the exception.

`econ_indicator`, `long_horizon` and `informed_flow_pillar` already carried the guard. Six sites did not, and the fix is no longer a copied predicate — it is one function, `broker.execution_gateway.booked_as_position`, beside `_FILLED_STATUSES`, which **the gateway's own `_record_result` now calls too**. Pillar bookkeeping and the gateway's `fills`/`cost_basis` writes therefore agree by construction rather than by nine hand-copied expressions staying in sync.

| pillar | why it bites |
|---|---|
| `bias_harvest` | `maker_entry` prices at the bid, so **every** order defers |
| `agent_trader` | corrupts the per-model P&L comparison its own intelligence-cap A/B exists to measure |
| `entailment_arb` | both legs written full-size; worse on the leg-B-failed path, where `submit_paired` had already cancelled the still-pending leg A |
| `term_structure`, `weather_temp`, `resolution_lens`, `vol_anchor` | byte-identical shape, unguarded |
| `cross_venue_arb` (×2) | the literal twin of `entailment_arb`'s two paths — **but** its `_record_leg` writes only a `signals` row, no `portfolio` row, so its phantom was a false traded-leg record rather than a phantom holding |
| `oddlot_tender` | `_record_entry` falls back to `float(qty)`, and nothing corrects it — `broker/sync.py` and `broker/reconciler.py` have no IBKR path. The phantom notional inflates `equity` in `risk/manager.py`, raising the 2%-of-equity stake for *every other market*. |

`resolution_tracker._settle_position` prefers the portfolio row over `cost_basis`, so these phantoms settle and book fabricated realized P&L.

**Severity, corrected.** The original draft of this PR said bias_harvest's paper record was "near-perfect … manufactured" from orders that never filled. That claim does not survive the data:

- bias_harvest has **201** `filled` trades rows (not 162), **199 of them before 2026-07-28** and the last two *on* 2026-07-28 — i.e. all of them at or before the day `paper_defer_resting_fills` landed. **All 201 have a matching `fills` row with `size > 0`** (minimum fill size 10.31, minimum price 0.09). The guard would not have blocked a single one.
- It has exactly **2 pending orders, a small total**, and **both later trade-through filled** — their `cost_basis` and `portfolio` rows agree exactly (a partial size at its limit price each).

So no fabricated paper record was manufactured for bias_harvest. What the guard prevents is a *future* class of phantom, and it prevents it across all nine sites. The measured standing damage today is small: 4 phantom `portfolio` rows against pending orders (`weather_temp` ×2, `term_structure` ×1, `resolution_lens` ×1, a small sum) plus the `agent_trader` rows the first draft already covered.

## 3. The mirror-image hole the guard opens

The guard's original comment claimed "the next `sync_positions` venue snapshot materializes the portfolio row." **That is false for paper.** `_record_deferred_paper_fills` calls `record_fill`, which writes `fills` + `cost_basis` and *nothing else*; position sync is mode-scoped (`is_paper_flag = 0 if settings.is_live else 1`), so a live bot never maintains paper rows — exactly what every `_record_position` docstring says.

Measured on the three pillars that adopted the guard earlier (2026-08-06):

- **`long_horizon`: 13 paper `cost_basis` positions, a material sum, with no `portfolio` row.**
- **`llm`: 18 more, a larger sum.**

They settle correctly (`check_resolutions` scans `cost_basis`) but are invisible to `RiskManager`'s position reads and to every pillar's `_open_position_count`, both of which read `FROM portfolio` — so `max_open` undercounts by that much.

`_record_deferred_paper_fills` now projects the portfolio row from `cost_basis` after `record_fill`. From `cost_basis` rather than from the single fill, deliberately: `cost_basis` carries the **cumulative** size and weighted-average cost, and the portfolio upsert *replaces* `size` rather than adding to it, so a fill-shaped write would silently undercount a market's second deferred fill.

**It cannot double-book.** `_settle_position` reads the portfolio row *or* the `cost_basis` row for `(size, entry_price)` — but it derives its idempotency key from neither: `source_ref = f"settle:{market_id}:{canon_token}:{is_paper_flag}"` comes from the position **key** alone, and `_prior_settlement` dedupes on it. Materializing the row therefore changes only *which branch* supplies the numbers, and since both branches now read the same `cost_basis` values, they supply identical numbers.

## 4. Known limitation: oddlot's live path

The resting branch's comment claimed it kept the filing out of the permanent dedupe "so a later cycle can still observe the fill." Also false: `run_once` dedupes on **row existence**, and `_audit_filing` already `INSERT OR IGNORE`d the accession *before* `_on_opportunity` runs; nothing anywhere reads `'order_resting'`.

So on the live path (`place_share_order` → `pending`, `filled_size=0`) a real 99-share fill is never booked. Giving it a real fill-observation route means a persisted order-id → accession mapping plus an `ib_async` order-status poll that calls `_record_entry` — a new order-lifecycle path for the equity book, and out of scope for a booking guard. **The comment now states the limitation plainly**, and the branch logs a `WARNING` rather than dropping silently. Latent twice over today: `oddlot_tender.paper: true` forces `dry_run` (which returns `status="paper"` with the full quantity, so this branch is not taken), and `ibkr.readonly: true` refuses live placement outright.

## Tests

`tests/test_fill_booking_guards.py` — 18 tests, **every one behavioural**. The previous revision had five tests that used `inspect.getsource` and asserted a string literal. Those were proven vacuous: replacing bias_harvest's guard with a computed-but-unused `_really_filled` followed by an unconditional record fully restored the regression **with all seven tests still green**.

Each replacement drives the real code path against an in-memory DB and asserts on rows: a stubbed exchange returning `pending`/`filled_size=0` must leave `portfolio` empty; the same path returning `paper`/`filled_size>0` must leave exactly one row whose size and price match the recorded fill. Coverage spans bias_harvest, weather_temp (newly swept), the cross_venue_arb leg path, oddlot's raw-`OrderResult` shape, both order-monitor branches, the real `PaperTrader` against a a nearly-empty wallet, and the predicate's own truth table across all seven order statuses.

**Eleven mutations were each verified to kill their test** — including the exact `_really_filled` mutation above, dropping the `filled_size` half of the predicate, removing the portfolio materialization, projecting it from the fill instead of `cost_basis`, and re-introducing a hand-rolled copy of the predicate in a tenth file.

- Full suite: **2258 passed, 5 skipped, 0 failed**
- Coverage: **66.94%** (floor 65.0)
- `ruff check .` — clean

### DIFF
--- a/auramaur/bot_order_monitor.py
+++ b/auramaur/bot_order_monitor.py
@@ -42,6 +42,17 @@ class OrderMonitorMixin:
         if db is None:
             return
         for result, order in filled:
+            # Same guard the LIVE branch applies below (`result.status ==
+            # "filled" and ... result.filled_size > 0`). The paper branch had
+            # none, so anything the paper trader handed over was booked
+            # verbatim — including a refusal stamped "filled". Defence in
+            # depth now that paper.py no longer produces those: a size-0 fill
+            # is never a fill, whichever book it came from.
+            if result.status != "filled" or result.filled_size <= 0:
+                log.debug("order_monitor.deferred_fill_skipped",
+                          order_id=result.order_id, status=result.status,
+                          filled_size=result.filled_size)
+                continue
             try:
                 if tracker is not None:
                     await tracker.record_fill(Fill(
--- a/auramaur/bot_order_monitor.py
+++ b/auramaur/bot_order_monitor.py
@@ -34,6 +34,7 @@ class OrderMonitorMixin:
         Fails soft: a bookkeeping error here must never take down the monitor
         loop that also reaps live orders.
         """
+        from auramaur.broker.execution_gateway import booked_as_position
         from auramaur.exchange.models import Fill
         from auramaur.research.polymarket_strategies import DecisionTracker
 
@@ -42,13 +43,14 @@ class OrderMonitorMixin:
         if db is None:
             return
         for result, order in filled:
-            # Same guard the LIVE branch applies below (`result.status ==
-            # "filled" and ... result.filled_size > 0`). The paper branch had
-            # none, so anything the paper trader handed over was booked
-            # verbatim — including a refusal stamped "filled". Defence in
-            # depth now that paper.py no longer produces those: a size-0 fill
-            # is never a fill, whichever book it came from.
-            if result.status != "filled" or result.filled_size <= 0:
+            # The same predicate the LIVE branch below applies, and the same
+            # one the gateway applies before IT writes a fill — see
+            # broker.execution_gateway.booked_as_position. The paper branch
+            # had none, so anything the paper trader handed over was booked
+            # verbatim, including a refusal stamped "filled". Defence in depth
+            # now that paper.py no longer produces those: a size-0 fill is
+            # never a fill, whichever book it came from.
+            if not booked_as_position(result):
                 log.debug("order_monitor.deferred_fill_skipped",
                           order_id=result.order_id, status=result.status,
                           filled_size=result.filled_size)
@@ -60,6 +62,7 @@ class OrderMonitorMixin:
                         token_id=order.token_id, side=order.side,
                         size=result.filled_size, price=result.filled_price,
                         fee=0.0, is_paper=True, order_id=result.order_id))
+                    await self._materialize_paper_portfolio_row(db, order)
                 if order.decision_id is not None:
                     await DecisionTracker(db).mark_fill(
                         int(order.decision_id),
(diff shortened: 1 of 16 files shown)


DarriEy:
Review (Claude, 2026-08-06): **NEEDS-AMENDMENT.** The defect class is real and I found live phantom rows proving it — but as written this PR fixes **3 of the 10 phantoms currently in the DB** and leaves the identical bug in five more files, while opening the mirror-image hole.

**Phantom rows in the live DB right now: 10.** Each has a `PAPER-LMT-*` trades row still `pending` and zero fills on the market. Only three are in this PR's blast radius; the other seven belong to pillars it does not touch — term_structure x2, resolution_lens, resolution_lens_kalshi x2, weather_temp x2. Plus **30 paper settlement rows on markets with zero fills, net negative**; the subset directly traceable to this exact defect (post-07-29 `PAPER-LMT-*`) is **15 rows** (agent_trader and weather_temp).

**Claim 2 (four pillars write a full-size portfolio row on `pending`) is REAL and measured.** `fill_size = result.filled_size if result.filled_size > 0 else order.size` turns a live `pending, filled_size=0` into a full-size row at the limit price. The gateway writes no cost_basis for pending, which is why portfolio-without-cost_basis is the signature; `_settle_position` reads portfolio first, so phantoms settle and book fabricated P&L, and `PortfolioTracker.get_positions` reads portfolio too, so they inflate RiskManager exposure.

**Claim 1 (paper.py stamps a refused order "filled") is real but has never fired.** Zero fills rows with size<=0 or price<=0; zero `trade_through` snapshots with filled_price<=0. Paper spendable sits far above the per-order stake — the balance gate has never been within two orders of magnitude of binding. Worth fixing as defence in depth; the body's "orders the paper trader refused counted toward promoting a strategy to live capital" describes a mechanism with no instances.

**The severity rhetoric does not survive the DB.** bias_harvest's 162 `filled` trades rows all predate 2026-07-28 (when paper_defer_resting_fills landed) and were genuine executes with filled_size>0 — the guard would not have blocked one. It has exactly 2 pending orders and **both trade-through filled**. No "near-perfect paper record" was manufactured. entailment_arb: 4 trades, all filled. oddlot: 0 entered. The guards are still correct; the claims need softening.

**Amendment 1 (load-bearing) — the sweep is incomplete.** Five more sites are byte-identical in shape and unguarded: `term_structure.py:998`, `weather_temp.py:149`, `resolution_lens.py:800`, `vol_anchor.py:423`, and `cross_venue_arb.py:398`+`:410` — the last being the literal twin of entailment_arb, same ok_statuses tuple, same `_record_leg`, same leg-B-failed path. Four are producing phantoms right now. Better than five more copies: factor the predicate into one shared helper (`booked_as_position(res)` in execution_gateway.py) so the class cannot reopen file by file.

**Amendment 2 — the fix opens the mirror-image hole.** The new comment claims "the next sync_positions venue snapshot materializes the portfolio row"; that is false for paper (`_record_deferred_paper_fills` writes only fills/cost_basis), and `_record_position`'s own docstring says the opposite. Measured in the three pillars that adopted this guard first: **long_horizon holds 13 paper cost_basis positions with no portfolio row at all** (llm another 16). They settle fine, but they are invisible to RiskManager and to `_open_position_count()`, so `max_open` undercounts. Write the portfolio row on the deferred paper fill.

**Amendment 3 — tests are mutation-proven vacuous.** The body's "6 of 7 fail against unfixed code" is accurate, but only 2 tests touch production code; the other 5 assert string literals via `inspect.getsource`. Proof: I replaced bias_harvest's guard with a computed-but-unused `_really_filled` followed by an unconditional record — the regression fully restored, every pending order writing a portfolio row again — and **all 7 tests still passed**, as did the whole related suite. They also break on any honest refactor (extracting a helper), so they pin the text and not the behavior. Replace with behavioral tests: stub the gateway returning pending/filled_size=0, assert zero portfolio rows; then paper/filled_size>0, assert one.

**Amendment 4 — oddlot's live path now loses real fills.** The new `else` says "Keep the filing out of the permanent dedupe … so a later cycle can still observe the fill", but `run_once` dedupes on row existence and `_audit_filing` already INSERT OR IGNOREd the accession before `_on_opportunity` runs. Nothing reads `'order_resting'`. So on the live path a real 99-share fill is never booked anywhere. Latent only because `oddlot_tender.paper: true` — a trap for whoever flips it.

**Amendment 5 — two false comments**: the bias_harvest one above, and paper.py's "cash may free up next tick" (`bot_order_monitor:134` calls `cancel_expired` in the same loop iteration, and `PaperTrader.cancel_expired` ignores its ttl argument and clears everything).

**Status semantics: correct.** `res.status != "pending" and filled_size > 0` is exactly the gateway's own `_FILLED_STATUSES and filled_size > 0` given the preceding early return. Partial fills are safe (they book `result.filled_size`, not full size). Every sentinel is already excluded rather than double-handled.

**Interaction checks all clean:** manual_trades (6e8c630) keys on live fills for SELLs and is untouched by a paper entry-BUY change; `reconciled_token`/`_prior_settlement` (3f70692) unaffected — this touches no token label, and the mirror gap cannot double-book because the cost_basis fallback reuses the same source_ref; #416's `cancel_resting` never reaches these guards, and at the one meeting point (submit_paired leg-A unwind) the PR is right — a cancelled pending leg A is correctly not recorded, while a genuinely filled leg A still is.

**Clock verdict: SAFE** — no config change, so no strategy_version change and no holdout resets. Non-clock note: this materially reduces evidence ACCRUAL RATE for bias_harvest and agent_trader.

I will raise the 10 phantom rows and the 15 fabricated settlements with the operator as a separate repair, with backups, following the `pnl_ledger_doublebook_backup_20260805` precedent.

DarriEy:
Amendments applied (4beadc8), rebased onto current main, force-pushed with lease. **2258 passed, coverage 66.94% >= 65.0, ruff clean.** Branch verified: 16 files, both commits present.

**A1 — the shared predicate, and a finding that matters beyond this PR.** `booked_as_position(res)` now lives beside `_FILLED_STATUSES` in execution_gateway.py, and critically **the gateway own fill write and its mark_fill evidence stamp call it too** — so pillar and gateway agree by construction rather than by copy. 16 sites routed (the 5 named, the 4 the PR fixed, 3 pre-existing, both order-monitor branches, 2 gateway-internal).

The design note deserves highlighting: the first implementation used `isinstance(res, ExecutionResult)` and broke 17 existing tests, because the house test doubles set `.status` on the outer object and `.filled_size` on `.result`. An isinstance check answers "not a position" for every such double — **a guard that passes every test and only fires in production.** Attribute probing instead, with a mutation pinning the shape. That is exactly the failure mode this PR exists to prevent, caught in its own fix.

**A2 — better than what I asked for.** I said "upsert using `_record_position`'s shape"; the agent projects the row **from cost_basis instead**, and the reason is right: cost_basis carries cumulative size and weighted-average cost, while the portfolio upsert REPLACES size — so a fill-shaped write would silently undercount a market second deferred fill and leave the two settlement branches pricing different holdings. Projecting from cost_basis makes them identical by construction. The double-booking argument is also stronger than "the refs happen to match": `_settle_position` derives its source_ref from the position KEY alone, so the new row changes only which branch supplies numbers, never the idempotency. A SELL that zeroes cost_basis deletes the row rather than leaving a mirror phantom.

**A3 — 7 tests (5 vacuous) became 18 behavioural, and 11 mutations were each proven to kill their test** — including the review own proof (bias_harvest guard replaced by a computed-but-unused variable), the isinstance shape, and a hand-rolled predicate re-added in vol_anchor. That last one is the important guard: it means the class cannot reopen by someone pasting the check back in.

**A4 — stating the limitation was the right call.** A real fill-observation route needs a persisted order-id to accession map plus an order-status poll, which is a new order-lifecycle path for the equity book and disproportionate here. The branch now logs a WARNING instead of dropping silently, and the gap is documented. Latent twice over anyway (`oddlot_tender.paper: true` AND `ibkr.readonly: true`).

**Corrections to my review, all of which I accept:**
- `cross_venue_arb._record_leg` writes only a **signals** row, not a portfolio row — unlike its entailment twin. Same shape, lesser consequence: a false traded-leg record rather than a phantom holding. Still worth guarding.
- bias_harvest has **201** filled trades rows, not 162, and **199 predate 2026-07-28 with 2 on it** — not "all predate". The substance holds: all 201 carry a fills row with size > 0, so the guard would not have blocked one.
- llm orphan cost_basis count drifted to 18 (from 16) — same phenomenon, still accruing.
- `interim_manager:214` and `platform_consensus:399` share the ok_statuses shape but record no position, so the site list is complete.

**On the standing-phantom count:** the report cites 4 phantom portfolio rows. My own survey against the live DB now returns **zero** — the operator-authorized repair ran concurrently with this work, removing the phantom position rows and the settlements fabricated from them, with backups under `*_phantom_backup_20260806`. The 4 rows remaining are `cost_basis` without fills, which is the mirror population this PR A2 addresses rather than the phantom one.

MERGE-READY.

PR #413, opened 2026-08-06
Fix five ways the accounting misreported what happened

DarriEy:
Five defects across four files, every one feeding a decision about capital —
plus the sibling of one of them on the promotion path, corrected in the second
commit after review.

## 1. `attribution.py` — token-blind `cost_basis` joins (4 sites)

`portfolio` and `cost_basis` are **both** keyed `(market_id, is_paper, token)`. Four joins matched only two of the three, so any market held on both sides fans out:

| | truth | reported |
|---|---|---|
| positions | 2 | **4** |
| exposure | a portion | **the remainder** |
| unrealized | the stake | **the gross figure** |
| realized (`sold_rows`) | the true value | **twice that** |

That shape is structurally normal for `entailment_arb`'s both-or-nothing legs and paired arbitrage. `cockpit.py:126` documents this exact bug and fixes it there — found via the web UI's duplicate-key warning. These were the siblings that never got it.

Every row of that table is now asserted by a test that builds the fixture and runs the query.

## 2. `ledger_report.py:149` — fees subtracted twice

`pnl_ledger.pnl` is **already net of fees** at every writer (`pnl.py:137`, `ledger.py:299`, `kalshi_settlements:152`, `instrument_booking:127`, `manual_trades`, `resolution_tracker`); the `fees` column is the breakdown of what's already deducted. `readiness.check_pnl_after_fees` states the convention outright — it comments `# already net of fees` on `SUM(pnl)` and then recovers gross as `net + fees`.

The risk-free benchmark computed `SUM(pnl) - SUM(fees)`, so a book at +the gross with the fee of fees printed **"net of fees a materially different number … behind cash"** in bold red, two lines under a `the gross figure` header.

### 2b. The same defect on the money path (added after review)

Correcting the display panel while `risk/graduation.py` (5 sites, the paper→live **promotion** gate) and `treasury/kraken_pillar.py` kept `SUM(pnl - fees)` would have left two contradictory conventions in-tree — the exact failure mode this PR is about. Both are corrected here.

Measured on the live DB first, because this **loosens** a promotion gate:

* Every strategy in the ledger carries `fees = 0.0` **except `kraken_directional`** — 26 live rows carrying the ledger's entire fee total, which is the entire ledger's fee total.
* `kraken_directional` has **zero** `decision_snapshots` and **zero** `strategy_experiments` rows, so it never reaches `_prospective_stats` — the only path `graduation.prospective_only=true` actually uses.
* Exactly **1 of 127** `report()` cells changes: `kraken_directional/kraken_spot`, −the overstated figure → −the true figure. Negative under both conventions. **No verdict flips.**

The Kraken desk is where it cost real dollars today: `_realised_pnl_usd` reported **−the overstated figure** against a true **−the true figure**.

`test_graduation_uses_pnl_net_of_fees` asserted the defect as the contract, on a fixture no writer can produce (`pnl=+1` alongside `fees=2`, when a 2 units fee on a 1 unit gross books `pnl=-1`). Rewritten.

## 3. `execution_gateway.submit_paired` — both decision ids discarded

`order.decision_id` stayed unset, so the `mark_fill` block in `_place_and_record` never ran and any paired-arb snapshot this path wrote would stay `filled=0` forever. With `graduation.require_executable_fills` true, `_prospective_stats` appends `AND d.filled = 1 AND d.fill_evidence IN (...)` — so `cross_venue_arb` and `entailment_arb` could never graduate on merit however long they ran.

**Scope of the damage so far, measured rather than assumed:** **zero** `decision_snapshots` rows and **zero** `strategy_experiments` rows for either strategy. No holdout clock was ever registered and no snapshot was ever written, so nothing was corrupted and nothing was burned. The first draft of this description claimed "every paired-arb snapshot stayed `filled=0` forever" and that the strategies were "still burning the 14-day holdout clock" — the first describes an empty set and the second is simply false. What this fixes is a gate these two strategies **could not have passed** had they started producing evidence, not evidence already spoiled.

`submit()` at `:136` plumbs these correctly. This path didn't.

## 4. `resolution_tracker.py` — settled one leg, deleted both

With `token_scope=None` the SELECT reads **one** row via `fetchone`, but the DELETE removed **every** token row for the market. A YES+NO holding booked one leg's P&L and dropped the other outright — the winning side never settled and never could, because its portfolio row was gone while its `cost_basis` row survived at `size > 0`. That day's `daily_stats` recorded the loss alone, which then feeds `check_daily_loss` against the the configured limit limit.

The `cost_basis` UPDATE **ten lines above** already scopes by the settled row's `token`, with a comment explaining why. The DELETE used `token_scope`.

The scoped DELETE is now **unconditional**. The first draft guarded it with `if token:`, but `token` is normalised to `(token or "YES").upper()` 100 lines earlier, so it is unconditionally truthy — the `else` arm was the whole-market DELETE this fix exists to remove, and it was unreachable.

## 5. `etf_calibration.py` — the trading gate scored against a coin

`Forecast.reference` defaulted to `0.5`, and `ibkr_etf_paper` passed `0.5` explicitly with this comment:

> fall back to a coin -- which **UNDERSTATES** the edge and therefore keeps the gate shut rather than opening it by accident.

It's backwards. `E[Brier]` of a constant forecast `r` under base rate `q` is `(r−q)² + q(1−q)`, so a coin benchmark is **easier** to beat than the instrument's own drift by exactly `(0.5−q)²`:

```
benchmark Brier vs COIN 0.5 : 0.25000
benchmark Brier vs drift 0.56: 0.24640
inflation of brier_edge     : +0.00360   <- free edge, every forecast
```

**How fast that opens the gate, corrected.** A flat forecaster answering 0.56 on everything — zero information, since 0.56 *is* the up-rate — does **not** clear at ~400 resolutions, as the first draft of this description claimed. At n=400 its 95% lower bound is still **−0.00224**; it first clears at **n=1026**. The edge is exactly 0.0036 at every n, so the defect is patience rather than speed — roughly two months at ~125 scoreable calls a week, against a `min_resolved` of 100. The constant and the n=400 bound are now pinned by test rather than asserted in prose.

There's no benchmark column on `ibkr_etf_forecasts`, so `reference` is now `None` and `clearance()` treats such forecasts as **unscoreable and refuses**. That's the honest state until the benchmark is recorded per forecast from `risk/ibkr_math.horizon_up_rate`.

### 5b. Keeping the kill signal audible (added after review)

`ibkr_etf_paper._clearance` designates the 5-session resolutions as *"an early kill signal: if the model shows no skill there within a few weeks, abandon"*. Passing `reference=None` is right for the **gate**, but the reporting path built its `Forecast`s the same way, so `auramaur ibkr-calibration` printed `LOCKED — N resolved forecasts carry no benchmark` and stopped. Silence is the one answer a kill signal must never give.

The report now prints the coin-referenced Brier edge as a separate, explicitly **kill-only** diagnostic beside the gate verdict. The asymmetry is what licenses it: a coin is the *easier* benchmark, so failing it is conclusive while passing it proves nothing — and the line says so. Against the live DB, read-only:

```
luna      103 resolved over 5s, Brier edge -0.00730 (95% low -0.01301)  NO SKILL
sol       101 resolved over 5s, Brier edge -0.00141 (95% low -0.00597)  NO SKILL
terra     102 resolved over 5s, Brier edge -0.00392 (95% low -0.00826)  NO SKILL
control     8 resolved over 5s, Brier edge -0.09000 (95% low -0.23346)  NO SKILL
```

**No arm has ever cleared on real data**, against a coin or otherwise. Read alongside the hit-rate table, which disagrees — luna calls 67.0% of moves right with a 57.2% lower bound. Direction and probability quality are different things; the trade gate is the Brier edge, and the diagnostic says which is which so the two are not pooled.

**The gate change is a no-op today.** All 103/101/102 resolved forecasts are 5-session. The traded 42-session horizon — the only one `_clearance` reads — has **0 resolved forecasts** across all four arms (175/169/173/79 rows, none due). The arm was locked before this PR and is locked after it, for a different and better-stated reason.

## Verification

`tests/test_accounting_scope.py` — 14 tests, plus a rewritten
`test_graduation_uses_pnl_net_of_fees`. Every behavioural claim is
mutation-verified: **8 mutants, one per fix site, each killed by the test that
names it.**

The first draft's fee test could not fail. It split the module source on
`risk_free_benchmark(`, which matches the `def` at line 21 rather than the call
at line 148, so it inspected the function **signature** —
`'*, net_pnl: float, first_at, last_at,\n settings=None'` — and passed
identically against pre- and post-fix source. Fixes 1 and 4 were covered the
same way, by `inspect.getsource` string counts executing zero lines of SQL.
They are now DB-backed.

CI-identical, in-container:

- **2256 passed, 5 skipped, 1 deselected, 0 failed**
- coverage **66.95%** (floor 65.0)
- `ruff check .` — clean

No file under `config/` is touched.

## Follow-up

Recording the real benchmark on `ibkr_etf_forecasts` is a schema change plus a
backfill and is deliberately not in this PR. Until it lands the ETF arm cannot
trade — which is the correct direction to fail, and the kill-signal diagnostic
means the decision to abandon can still be made from evidence in the meantime.

### DIFF
--- a/auramaur/broker/execution_gateway.py
+++ b/auramaur/broker/execution_gateway.py
@@ -479,13 +479,25 @@ class ExecutionGateway:
                                 reason=cap_b or "paired leg blocked by market cap"),
             )
 
-        await self._capture_decision(a, order_a)
-        await self._capture_decision(b, order_b)
+        # Keep the ids. Discarding them left order.decision_id unset, so the
+        # mark_fill block in _place_and_record never ran and EVERY paired-arb
+        # snapshot stayed filled=0 forever. With
+        # graduation.require_executable_fills true (the tracked YAML),
+        # _prospective_stats appends `AND d.filled = 1 AND d.fill_evidence IN
+        # (...)`, so cross_venue_arb and entailment_arb could never graduate
+        # paper->live on merit — while still burning the 14-day holdout clock
+        # and writing snapshots that looked like accumulating evidence.
+        # submit() at :136 plumbs these correctly; this path did not.
+        decision_id_a = await self._capture_decision(a, order_a)
+        decision_id_b = await self._capture_decision(b, order_b)
+        order_a.decision_id = decision_id_a
+        order_b.decision_id = decision_id_b
 
         res_a = await self._place_and_record(
             order_a, strategy_source=a.signal.strategy_source,
             signal_id=getattr(a.signal, "id", None),
-            exchange=exchange_a, exchange_name=exchange_name_a)
+            exchange=exchange_a, exchange_name=exchange_name_a,
+            decision_id=decision_id_a)
         if res_a.status not in _OK_STATUSES:
             # Leg A rejected — B is never placed, nothing to unwind.
             return res_a, ExecutionResult(status="skipped", reason="leg_a_not_ok")
@@ -493,7 +505,8 @@ class ExecutionGateway:
         res_b = await self._place_and_record(
             order_b, strategy_source=b.signal.strategy_source,
             signal_id=getattr(b.signal, "id", None),
-            exchange=exchange_b, exchange_name=exchange_name_b)
+            exchange=exchange_b, exchange_name=exchange_name_b,
+            decision_id=decision_id_b)
         if res_b.status not in _OK_STATUSES:
             # Leg risk: A is in, B failed. Cancel a live-pending A so we don't
             # sit on a naked directional leg (paper / already-filled A can't be
--- a/auramaur/broker/execution_gateway.py
+++ b/auramaur/broker/execution_gateway.py
@@ -480,13 +480,20 @@ class ExecutionGateway:
             )
 
         # Keep the ids. Discarding them left order.decision_id unset, so the
-        # mark_fill block in _place_and_record never ran and EVERY paired-arb
-        # snapshot stayed filled=0 forever. With
+        # mark_fill block in _place_and_record never ran and any paired-arb
+        # snapshot this path wrote would stay filled=0 forever. With
         # graduation.require_executable_fills true (the tracked YAML),
         # _prospective_stats appends `AND d.filled = 1 AND d.fill_evidence IN
         # (...)`, so cross_venue_arb and entailment_arb could never graduate
-        # paper->live on merit — while still burning the 14-day holdout clock
-        # and writing snapshots that looked like accumulating evidence.
+        # paper->live on merit however long they ran.
+        #
+        # Scope of the damage so far, measured rather than assumed (live DB,
+        # 2026-08-06): ZERO decision_snapshots rows and ZERO
+        # strategy_experiments rows for either strategy. No holdout clock was
+        # ever registered and no snapshot was ever written, so nothing was
+        # corrupted and nothing was burned — the defect was a gate these two
+        # strategies could not have passed had they started producing
+        # evidence, not evidence already spoiled. Fixed before that mattered.
         # submit() at :136 plumbs these correctly; this path did not.
         decision_id_a = await self._capture_decision(a, order_a)
         decision_id_b = await self._capture_decision(b, order_b)
--- a/auramaur/cli/reporting.py
+++ b/auramaur/cli/reporting.py
@@ -187,6 +187,8 @@ def ibkr_calibration(arm: str, width: float):
     coin flip until proven otherwise.
     """
     async def _run():
+        from dataclasses import replace
+
         from auramaur.evaluation.etf_calibration import (
             COIN, CONFIDENCE_RANKS, Forecast, brier_score, clearance,
             confidence_bands, probability_bands, samples_for_reliable_edge,
@@ -206,7 +208,7 @@ def ibkr_calibration(arm: str, width: float):
                 clause, params = " WHERE model_alias = ?", [arm]
             rows = await db.fetchall(
                 f"""SELECT model_alias, probability, confidence, actual_outcome,
-                           due_at
+                           due_at, horizon_sessions
                     FROM ibkr_etf_forecasts{clause}
                     ORDER BY model_alias, id""", tuple(params))
             if not rows:
@@ -214,14 +216,23 @@ def ibkr_calibration(arm: str, width: float):
                 return
             by_arm: dict[str, list] = {}
             pending: dict[str, list[str]] = {}
+            resolved_horizons: dict[str, set[int]] = {}
             for row in rows:
                 by_arm.setdefault(row["model_alias"], []).append(Forecast(
                     float(row["probability"]), str(row["confidence"]),
                     None if row["actual_outcome"] is None
-                    else int(row["actual_outcome"])))
+                    else int(row["actual_outcome"]),
+                    # No benchmark column on ibkr_etf_forecasts, so these are
+                    # unscoreable to the gate. The COIN diagnostic below is
+                    # what keeps the kill signal readable meanwhile.
+                    None))
                 if row["actual_outcome"] is None:
                     pending.setdefault(row["model_alias"], []).append(
                         str(row["due_at"]))
+                else:
+                    resolved_horizons.setdefault(
+                        row["model_alias"], set()).add(
+                            int(row["horizon_sessions"] or 0))
 
             for alias in sorted(by_arm):
                 forecasts = by_arm[alias]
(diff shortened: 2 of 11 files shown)


DarriEy:
Review (Claude, 2026-08-06): **NEEDS-AMENDMENT.** The body says four defects; there are five, and I judged all five. All are directionally right, the suite is green, and — importantly — **no fix implies any historical repair scope**. Everything is either read-time or forward-looking. The amendments are about an untested fix, a self-contradicting comment, an unswept money-path sibling, and a muted kill signal.

**Fix 1 (attribution.py token-blind joins) — CORRECT, ship as-is.** Four joins matched `(market_id, is_paper)` on tables keyed `(market_id, is_paper, token)`. cockpit.py and web/queries.py already carry the fix in the identical form, so this completes the sweep — no unscoped siblings remain. Measured against the live DB on the markets held on both sides: the live position count is overstated by **~12%** and live exposure by **~10%**, with the polymarket venue line overstated 18.1%. Note `news_speed` drops 38% while `market_maker` goes UP — proving cross-token cost-basis contamination rather than simple doubling. The body's "realized $14->$28" row is structurally right but has never fired: every both-sided market carries realized_pnl = 0.

**Fix 2 (fees subtracted twice) — right premise, INCOMPLETE and UNTESTED.** Confirmed at all six writers that `pnl` is net and `fees` is the breakdown. Real magnitude: the live panel understated net P&L by **7.7%**. Every recorded fee belongs to kraken_directional; paper fees are zero, so the paper panel was never wrong. The body's "+$40 book with $55 fees" scenario does not exist. Three problems: (a) `ledger_report.py:216-219` still documents *"this annualises SUM(pnl - fees)"* — the PR deleted that behavior and left the comment, so the file self-contradicts; (b) **the load-bearing sibling is untouched** — `risk/graduation.py` uses SUM(pnl - fees) at :100, :110, :156, :285, :287, which is the paper->live promotion path, plus `kraken_pillar.py:1138`. Shipping the display fix while promotion keeps the old convention is the worst of both; (c) its test is **vacuous** — see below.

**Fix 3 (submit_paired dropped decision ids) — CORRECT code, consequence unsupported.** The divergence from `submit()` is real and worth fixing. But the DB shows the harm never occurred: **zero decision_snapshots for either arb strategy** (of 481) and zero strategy_experiments rows. entailment_arb's only 4 trades predate snapshot data entirely. So "every paired-arb snapshot stayed filled=0 forever" describes an empty set, and "still burning the 14-day holdout clock" is false — no clock was ever registered. Soften the body.

**Fix 4 (settled one leg, deleted both) — CORRECT and the most valuable, severity overstated.** The unscoped DELETE is the main path and the fix is right. But "never settled and could never settle" is wrong: the cost_basis arm of the union re-lists the market next cycle and settles the survivor under a different canon_token — the PR's own new comment concedes this, and the DB agrees (28 markets settled both sides historically; **zero** orphan cost_basis rows; **zero** portfolio rows on settled markets). Historical repair scope is effectively zero. The value is prospective: **markets are held on both sides right now** and every one will traverse this path at resolution. One defect in the fix: `token = (token or "YES").upper()` makes `token` unconditionally truthy, so the `else` at :748-751 is unreachable dead code.

**Fix 5 (etf coin benchmark) — reasoning right, headline number FALSE, and it costs a documented kill signal.** The math checks out: at q=0.56, coin 0.25000 vs drift 0.24640 = +0.00360 free edge, so the old comment was genuinely backwards. But the body's "clears at n~400" is wrong — a flat q=0.56 forecaster has lo = **-0.00224 at n=400** and first clears at **n=1026**. And it never happened: recomputing pre-PR `clearance()` on the real rows gives **luna lo -0.013006, sol -0.005967, terra -0.008260, all cleared=False**. The gate was already shut for a different reason — `_clearance` filters on the traded 42-session horizon, which has **0 resolved forecasts**, so with an empty list the new early-return never fires and the change is a **no-op on the gate today**. Where it DOES change behavior is a regression: `cli/reporting.py:218` builds Forecast with no reference, so `auramaur ibkr-calibration` will print "LOCKED - no benchmark" instead of the measured Brier edge. `ibkr_etf_paper.py:453-457` explicitly designates those 5-session numbers as **"an early kill signal: if the model shows no skill there within a few weeks, abandon"** — and the signal currently says kill. This PR mutes it. Also: the early-return is `if unscoreable and not resolved`, so on a MIXED population unscoreable rows are silently dropped and the gate is judged on the scoreable subset — selection bias, untested.

**On the is_holdout observation in the review notes — I checked it and it is NOT a defect.** 481 snapshots all stamped 0, yes, but `is_holdout` is stamped at insert by comparing against `holdout_starts_at`, and **zero snapshots were mis-stamped after their window opened**. The stamping is correct; the zeros are the 14-day warmup working. The real finding is what the query showed instead: **37 registered versions, not one past warmup**, earliest opening 2026-08-07.

**Clock verdict: SAFE.** `git diff origin/main -- config/` is empty. **Collision check vs the two prior repairs: CLEAN** — no backfill or recompute added, backup tables untouched, the 2 manual-sell rows and all 5 deliberately-left ambiguous twins intact, `_prior_settlement` untouched so the 3f70692 dedup contract holds. Worth watching (not a re-break): fix 4 makes `_settle_position` re-enter more often, raising traffic on the exact path repaired yesterday — the dedup holds because canon_token differs per leg.

**Test quality — 3 of 7 weak, and the vacuous one is the ONLY test for fix 2.** `test_benchmark_does_not_subtract_fees_twice` splits on `"risk_free_benchmark("`, which hits the **def line** first, so it asserts against the function signature and passes identically pre- and post-PR. Proven empirically. `test_zero_skill_arm_no_longer_clears_against_a_coin` is misnamed — it passes reference=None, so it never scores against a coin, and pre-PR it fails with TypeError rather than a measurement.

**Amendments:** (1) fix the vacuous fees test — call `gather_ledger_report` against a temp DB with pnl=+40/fees=55 and assert net_pnl == 40.0; (2) update ledger_report.py:216-219, it documents deleted behavior; (3) decide graduation.py + kraken_pillar.py:1138 — fix or file the follow-up, do not leave promotion on the old convention silently; (4) restore the ETF kill signal in cli/reporting.py:218; (5) correct the body's claim-5 numbers and soften claim 3; (6) drop the unreachable else at resolution_tracker.py:748-751.

**Two findings for their own tickets:** all three ETF arms show negative Brier edge on ~100 resolved 5-session forecasts each — decision-relevant, and it is the signal this PR would hide. And no strategy version has yet reached its holdout window.

DarriEy:
Amendments applied (eab3762), rebased onto current main, force-pushed with lease. Verified: **2256 passed, coverage 66.95% >= 65.0, ruff clean.** Branch integrity confirmed — both commits present, 11 files.

**A3 decision — fixing both money-path sites here rather than deferring — is right, and the measurement supports it.** Every strategy in pnl_ledger carries fees = 0.0 except kraken_directional (26 rows carrying the ledger's entire fee total), and kraken_directional has zero decision_snapshots and zero strategy_experiments, so it never reaches `_prospective_stats`. Exactly 1 of 127 report cells changes and no verdict flips. Meanwhile `kraken_pillar._realised_pnl_usd` was reporting a materially wrong realized figure right now, so deferring would have shipped the display fix while leaving a wrong number on the desk. Good catch that `test_graduation_uses_pnl_net_of_fees` encoded the old arithmetic with a fixture no writer can produce (pnl=+1 alongside fees=2; a fee of 2 on gross of 1 books pnl=-1).

**A1 proven directly** — the original test passes against BOTH pre- and post-fix source because `split("risk_free_benchmark(")` matches the def line at 21, not the call at 148. Eight mutants raised, eight killed.

**One correction, and it reverses the amendment report's most interesting claim.** The report states the arms have real directional skill (luna "67.0% of moves right, 57.2% Wilson lower bound, beats coin: yes") and that only the Brier statistic says kill. **That number is not a hit rate.** `etf_calibration._band` computes `hits = sum(int(f.actual_outcome))` and `realized = hits / n` — the count of UP outcomes, never compared against the forecast. So the "hit rate" column is the market's base rate of upward moves, and "beats coin" tests that base rate against 0.5. Any arm operating in a market that rose 66% of the time shows "67%, beats coin: yes" even with pure-noise forecasts.

Recomputing actual directional accuracy (`(p > 0.5) == (outcome == 1)`) over the same resolved rows:

| arm | n | true hit rate | Wilson low | mean forecast | base rate (the "67%") |
|---|---|---|---|---|---|
| luna | 103 | **42.7%** | 33.6% | 0.496 | 0.660 |
| sol | 101 | **51.5%** | 41.9% | 0.501 | 0.663 |
| terra | 102 | **44.1%** | 34.9% | 0.498 | 0.667 |
| momentum_control | 8 | 37.5% | 13.7% | 0.500 | 0.375 |

All three forecast ~0.50 constantly while the market rose ~66% of the time. A trivial always-predict-up rule would score 66% — every arm is far below that and at or under chance. **The kill signal stands; there is no hidden directional skill.** The diagnostic this PR restores is therefore correctly framed, and its "NO SKILL" verdicts are right.

The mislabeling is a separate live defect and I am filing it: `confidence_bands`' docstring asks "does a higher stated confidence actually predict a higher hit rate?" and the code cannot answer that question. The mean_forecast-vs-realized comparison is a legitimate calibration curve; the "hit rate" and "beats coin" LABELS on it are misleading enough that they misled a careful reviewer into concluding these arms have skill.

PR #414, opened 2026-08-06
Make degraded input fail closed instead of reading as a benign value

DarriEy:
Five sinks that turned **"I could not measure this"** into a plausible number. Each one is individually small; together they're the reason a degraded system reads as a healthy one.

## 1. `kraken_pillar.py:710` — an empty balance dict read as a flat book

`get_free_balance` falls back to `get_balance`, which returns `{}` on **any** API error — `kraken.py:121-123` logs and returns the empty result rather than raising. One 5xx, rate-limit or `EAPI:Invalid nonce` made every tracked position look closed:

- `_dir_long` cleared
- `_clear_peak()` **DELETED** the `position_peaks` row — the trailing-stop high-water mark, which is **not** recoverable from `cost_basis`
- `_mirror_to_portfolio` removed every live kraken portfolio row

A position that peaked +40% and sat at +30% with an 8% trailing stop re-anchored at +30%; the give-back was never banked. `_treasury` guards this exact case at line 419 with `if not bal: return`.

## 2. `gates.py:21` — an unreadable database read as zero

The path was the bare literal `"auramaur.db"`, which resolves only under a CWD that happens to be the repo root. Under compose (`/app/state/auramaur.db`) or from anywhere else, the connect raised and `except: return 0` laundered it into `0/100 resolved-$ markets → WAIT (need data)` — so the operator never flips a filter the data has already earned.

Now: resolves via `runtime.db_path()`, sets `busy_timeout=5000` so a checkpoint in flight isn't the same failure, and returns `-1` rendered as **"UNKNOWN (cannot read db)"** — a genuinely different state from "no data yet".

## 3. `readiness.py:140` — an empty log window read as zero errors

The window can be empty because the bot is dead, **or because rotation carried the errors away**: `_rotate()` renames at `rotate_max_mb` (10MB, 3 backups) and this parser opens exactly one path, so a bot writing >10MB/day gives a criterion advertising 7 days under a day of actual visibility.

An error burst two rotations ago scored `PASS`. With the other criteria green, `overall_pass` was True and the CLI exited 0 — an automated *"ready to arm live"* built on evidence nobody could see. Now `INSUFFICIENT_DATA`, which `overall_pass` already treats as not-pass.

## 4. `tolerance.py:23` — the last CWD-relative state path, governing the whole risk surface

`Path("data/risk_tolerance")` wins over config and scales **Kelly ×3, min-edge ÷3, category cap → 100%, divergence max → 0.75, max open positions → 1500**.

A stale file in a launch directory silently put the book at maximum aggression with no log line. Conversely `auramaur risk 0` run from any directory other than the bot's had no effect at all — while printing that it did. `runtime.py` exists to anchor exactly this, and `grep 'Path("data/'` returned this line and nothing else.

## 5. `redeem.py:144` — `dust-exit` attached to a CWD-relative database

Its **only** concurrency protection is a flock on `f"{db_path}.lock"` — also CWD-relative. Under any deployment that relocates state it flocked an uncontended `./auramaur.db.lock`, so the documented *"refuses to run while the bot is running"* guard never engaged.

It then attached to `./auramaur.db`: either a stale pre-migration file — in which case it exits real positions the running bot is concurrently managing, the double-sell `tests/test_cli_dust_exit.py:110` exists to prevent — or a nonexistent one, which `assemble_components` **creates** with full schema DDL before reporting "No dust positions" to an operator who then concludes there are none.

## Verification

`tests/test_degraded_input_fails_closed.py` — 5 tests. **All 5 fail against the unfixed code.**

- Full suite: **2111 passed, 5 skipped, 0 failed**

### DIFF
--- a/auramaur/cli/redeem.py
+++ b/auramaur/cli/redeem.py
@@ -8,6 +8,7 @@ import click
 from rich.table import Table
 
 from auramaur.db.database import Database
+from auramaur.runtime import db_path as runtime_db_path
 from config.settings import Settings
 
 from auramaur.cli._base import console, main
@@ -141,7 +142,21 @@ def dust_exit(max_notional: float, exchange: str | None, execute: bool, yes: boo
 
         settings = Settings()
         from auramaur.cli import AuramaurBot  # call-time lookup keeps test patch working
-        bot = AuramaurBot(settings=settings, db_path="auramaur.db", exchange_filter=exchange)
+        # Resolve through runtime.db_path() the way run.py:61 does. The bare
+        # literal only resolves under a CWD that happens to be the repo root,
+        # and this command's ONLY concurrency protection is the flock on
+        # f"{db_path}.lock" — also CWD-relative. Under any deployment that
+        # relocates state (compose sets /app/state/auramaur.db; deploy/*.sh set
+        # AURAMAUR_DB_PATH) it flocked ./auramaur.db.lock, which is
+        # uncontended, so the "refuses to run while the bot is running" guard
+        # never engaged. It would then attach to ./auramaur.db — either a stale
+        # pre-migration file, in which case it exits real positions the running
+        # bot is concurrently managing, or a nonexistent one, in which case
+        # composition.assemble_components CREATES it with full schema DDL and
+        # reports "No dust positions" to an operator who then concludes there
+        # are none.
+        bot = AuramaurBot(settings=settings, db_path=str(runtime_db_path()),
+                          exchange_filter=exchange)
         try:
             await bot._init_components()
         except RuntimeError as e:
--- a/auramaur/monitoring/gates.py
+++ b/auramaur/monitoring/gates.py
@@ -11,27 +11,49 @@ from __future__ import annotations
 
 import sqlite3
 
+import structlog
+
 from rich.table import Table
 from rich.text import Text
+
+from auramaur.runtime import db_path as runtime_db_path
+
+log = structlog.get_logger()
 
 
 def _resolved_dollar_markets(db_path: str) -> int:
     """Count live markets with fills AND a resolved outcome (the $-edge sample)."""
+    # busy_timeout: without it a checkpoint in flight raises SQLITE_BUSY, which
+    # the except below turns into 0 — indistinguishable from "no data".
     try:
-        c = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True)
+        c = sqlite3.connect(f"file:{db_path}?mode=ro", uri=True, timeout=5.0)
+        c.execute("PRAGMA busy_timeout=5000")
         n = c.execute(
             "SELECT COUNT(DISTINCT f.market_id) FROM fills f "
             "JOIN calibration cal ON cal.market_id=f.market_id "
             "WHERE f.is_paper=0 AND cal.actual_outcome IS NOT NULL").fetchone()[0]
         c.close()
         return n
-    except Exception:
-        return 0
+    except Exception as exc:
+        # Returning 0 makes an unreadable database look like an empty one, and
+        # the gate dashboard renders that as "WAIT (need data)" — so an
+        # operator never flips a filter the data has already earned. Surface
+        # the failure instead of laundering it into a plausible number.
+        log.warning("gates.resolved_dollar_markets_unreadable",
+                    db_path=db_path, error=str(exc)[:160])
+        return -1
 
 
-def gather(settings, db_path: str = "auramaur.db") -> list[dict]:
+def gather(settings, db_path: str | None = None) -> list[dict]:
     from auramaur.risk.tolerance import current_tolerance
-    n_resolved = _resolved_dollar_markets(db_path)
+    # Resolve through runtime.db_path() so this follows AURAMAUR_DB_PATH /
+    # AURAMAUR_STATE_DIR the way Database.__init__ does. The old default was
+    # the bare literal "auramaur.db", which only resolves under a CWD that
+    # happens to be the repo root — under compose (/app/state/auramaur.db) or
+    # from any other directory the connect raised, the except returned 0, and
+    # the dashboard reported "0/100 resolved-$ markets -> WAIT (need data)"
+    # while hundreds existed.
+    n_resolved = _resolved_dollar_markets(db_path or str(runtime_db_path()))
     mc = settings.momentum_coupling
     tol = current_tolerance(settings)
     tol_label = ("most conservative" if tol <= 15 else "conservative" if tol < 45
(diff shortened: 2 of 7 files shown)


DarriEy:
Review (Claude, 2026-08-06): **NEEDS-AMENDMENT.** All five conversions are directionally right and none can refuse a cancel, exit, settlement or reconciliation in a way worse than the status quo — but three need work before merge, and two of the five are fixing defects that are ACTIVE in the running deployment right now.

**Verified live, not theoretical:**
- **gates.py**: the bare `"auramaur.db"` literal resolves only under a repo-root CWD, and `/app/auramaur.db` does not exist in the container — so `auramaur gates` prints `0/100 resolved-$ markets -> WAIT (need data)` while the true count is **above the threshold**. The gate has already earned GO and the dashboard has been saying otherwise.
- **redeem.py**: `dust-exit` in the container attached to that same nonexistent path, so `assemble_components` created an empty schema and it reported "No dust positions" — a **silent no-op**, while flocking an uncontended lock so the documented double-sell guard never engaged. Post-fix it either operates on the real book or refuses honestly. Worth saying in the body: operators who saw "No dust positions" in-container were being lied to.
- **kraken_pillar**: measured against 4 rotations of live logs (246k lines), `kraken.balance_error` fires **5 times against 865 successes — 0.58%, about once a day.** A real recurring transient.

**On that Kraken guard, the classification cuts the other way from #407 and that is the key finding.** The error class IS transient, but the guard is strictly EXIT-FAVORABLE: pre-change, `bal={}` made `_reconcile_positions` see every position as closed, so it popped `_dir_long`, **deleted the position_peaks row**, and wiped the portfolio mirror — then the loop ran with `holding=False` and attempted no sell either. The old code did not exit, it FORGOT, and destroyed the trailing high-water mark on the way out. The guard skips one cycle and keeps both, so the next healthy poll (<=5 min) exits at the true peak. That is the inverse of the #407 regression, not a repeat. The paper carve-out is also correctly reasoned — verified that paper position state comes from `kraken_paper_positions`, so without it a credential-less paper deployment would never trade again.

**Required amendments:**

**A1 — readiness.py: the code does not do what its comment claims.** The comment describes fixing rotation blindness ("an error burst two rotations ago scored PASS"); the code only catches a fully empty window. Measured: the current log covers **0.7 of 7 requested days (10%)**, rotating every ~21-30h, so the scenario the comment describes is still live and the new branch never fires (0% trip rate). Either parse the rotated backups so the 7-day criterion sees 7 days, or return the earliest well-formed timestamp `_parse_log_for_errors` already computes and discards, and report honest coverage. Be aware the second form trips on 100% of current data — honest, harmless (readiness gates nothing automatic), but permanently not-pass until rotation is followed. Parsing backups is the better fix. If neither, strike the rotation paragraph.

**A2 — kraken_pillar: escalate a sustained outage.** A revoked or expired key at 3am silences the entire exit ladder indefinitely behind a `log.warning`, and `warning` is not in readiness's `_ERROR_LEVELS`, so nothing escalates. Count consecutive empty-balance cycles, escalate to `log.error` after k=3 (15 minutes), reset on the first non-empty balance.

**A3 — tests: 4 of 5 are self-referential.** They use `inspect.getsource` and assert a string literal appears — they execute no production code and would pass against `if not bal and False: return`. **The transient case is untested everywhere**, which is exactly the half that went wrong in #407. Needed: `_directional` with `get_free_balance -> {}` in LIVE mode asserting `_dir_long` unchanged, the `position_peaks` row still present, and `place_spot_order` never called; the **paper carve-out still trading** on `bal={}` (non-negotiable — that is the availability half and it ships untested); `check_cycle_health` end-to-end across empty/clean/errored windows; and `gather()` actually rendering `UNKNOWN (cannot read db)`.

**Clock verdict: CLEAN.** No config/defaults.yaml or config/settings.py change. Confirmed independently that risk tolerance never enters `strategy_version` — the hash is built from raw `settings.risk`, and `scale_risk` returns a `model_copy` without mutating it.

Nits: `gates.py` `c.close()` is not in a `finally` (and #399 just landed a leaked-connection guard); `tolerance.py` freezes `_OVERRIDE` at import time, unlike every other runtime-path consumer which resolves at call time; and the diff deletes PEP8 blank lines between six top-level defs in a risk-surface file — unrelated churn, revert to keep the diff honest.

Amendments being applied now.

DarriEy:
Amendments applied (a874c5b) and the branch rebased onto current main (Claude, 2026-08-06). Note this rewrote the PR commit (a49a2de -> 4f01186) and force-pushed with --force-with-lease; both commits are present and the diff is 7 files. Container verification: **2243 passed, coverage 66.88% >= 65.0, ruff clean.**

**A1 — readiness now parses the rotated backups.** Measured against the real logs (auramaur.log + .1/.2/.3) through the patched production code:

| | entries in window | errors | coverage | verdict |
|---|---|---|---|---|
| before | 51,955 | 1,445 | 0.71d / 7d (10.1%) | FAIL |
| after | 247,055 | 2,932 | 3.90d / 7d (55.7%) | FAIL |

**Correction to my review, which the agent is right about:** I implied the rotation fix would change the criterion's verdict. It does not today — the active file already contains errors, so it was FAIL either way. What changed is that it now fails on evidence rather than by luck: **1,487 error events across the three backups were invisible to the old parser**, and had the active file been clean the criterion would have scored PASS on them. That is precisely the scenario the original comment described and the code did not implement. Rendered value is now honest about partial coverage.

**Worth knowing, and deliberately not fixed here:** full 7-day coverage is unreachable at the current write rate — 4 files x ~10MB spans only ~3.9 days, so the criterion will keep reporting `log covers ~3.9d of 7.0d` until `rotate_backups` is raised. That is a config change and therefore a graduation-clock reset, so it is an operator decision, not a drive-by.

**A2 — escalation.** Module constant at 3 consecutive empty reads (15 min at the 300s cycle), counter incremented on every empty read and reset only on a genuinely non-empty one, both mode-independent so a paper/live flip cannot mute a live alert part-way. Below threshold warns; at threshold `log.error("kraken.directional.balance_unavailable_sustained")`, which `_ERROR_LEVELS` catches. Paper carve-out preserved. No config knob, deliberately.

**A3 — 5 self-referential tests replaced by 22 executing ones, each mutation-verified.** Every replacement was checked by neutering the production fix and confirming the test fails. Highlights: `_directional()` LIVE on `bal={}` against a real DB asserting `_dir_long` unchanged, the `position_peaks` row intact at 40.0, and `place_spot_order` never called — plus a discriminator proving a *successful* read holding no position still clears both; the **paper carve-out now covered in both directions** (still exits via stop-loss, still enters), which was the untested availability half; `check_cycle_health` end-to-end across empty/clean/errored/rotated-away windows; and `dust-exit` through `CliRunner` asserting the `db_path` kwarg AuramaurBot actually receives.

Nits folded in: `contextlib.closing` on the gates connection, `_OVERRIDE` converted to a call-time `_override_path()` matching the house shape in `transfers.py`, and the PEP8 blank lines restored.

PR #415, opened 2026-08-06
Close the prompt-injection gaps in six trade-driving prompts (#405)

DarriEy:
Closes **#405 item 2** (zero-width / tag characters in the scrubber) and converts the six prompt sites catalogued in the #410 review comment on that issue. Item 1 (the deterministic pairing post-check) is untouched and remains open.

## Part 1 — the scrubber gap (#405 item 2)

`_CONTROL_OR_BIDI` stripped C0/C1 controls and BiDi overrides but not zero-width characters (U+200B–200D, U+FEFF), LRM/RLM, the word joiner, soft hyphen, or the Unicode TAG block (U+E0000–E007F). Verified empirically against this exact pipeline: NFKC leaves all of them, and none has the White_Space property, so `" ".join(text.split())` keeps them too. The tag block is a complete invisible ASCII alphabet — an instruction can reach the model spelled in text that renders as nothing in a terminal, a diff, or code review. The class now names all three families, and a parametrized test asserts each codepoint is gone.

**U+200D (ZWJ) is stripped deliberately.** It is load-bearing in emoji ZWJ sequences, so this is a real if small loss (a family emoji collapses into its parts). These prompts carry market questions, resolution criteria, news text and SEC filings to a model that reasons about them analytically; glyph fidelity buys nothing there, while an invisible joiner inside a delimiter-shaped payload costs something. The call is documented in a dated comment and pinned by a test rather than left implicit.

`format_untrusted_block()` (the body of `strategic._scrub`, which now delegates to it) moves into `nlp/prompts.py` so the pillars share one definition instead of importing a private name across packages.

## Part 2 — per-site conversion

| Site | What was raw | Now |
|---|---|---|
| `strategy/resolution_lens.py` (LENS / VERIFY / GROUND) | question, FULL criteria, open-web evidence lines, replayed mechanism | all scrubbed; one `<UNTRUSTED_MARKET_TEXT>` / `<UNTRUSTED_MARKET_AND_EVIDENCE>` pair per prompt |
| `strategy/agent_trader.py` (MANDATE, `_candidates_block`, memory, open book) | question raw (description had only `.replace("\n"," ")`), stored question+thesis raw | all scrubbed; one `<UNTRUSTED_TRADING_DATA>` pair around record + open book + candidates |
| `strategy/agent_analyzer.py` (`_format_markets_for_agent`, deep research) | id, question, description, category raw | all scrubbed; `<UNTRUSTED_MARKETS_BLOCK>` and `<UNTRUSTED_MARKET_BLOCK>`; deep-research prompt lifted to a module constant so its boundary is testable |
| `strategy/term_structure.py` (CURVE_PROMPT) | `family=question`, `rules=description[:1500]`, strike ids | all scrubbed; one `<UNTRUSTED_LADDER_BLOCK>` pair |
| `strategy/settlement_arb.py` (EXTRACT / VERIFY) | question + `description[:2000]` | scrubbed; one `<UNTRUSTED_MARKET_TEXT>` pair per prompt |
| `strategy/oddlot_tender.py` (ODDLOT_PROMPT) | company/ticker/form/date + `filing[:40000]` | scrubbed in place; one `<UNTRUSTED_FILING_BLOCK>` pair |

Highest-severity first: `resolution_lens` is both a trade gate (a forged verdict JSON sets `fair_prob`/`gap_score`, which is the whole entry decision) and the consumer of open-web evidence, the input an attacker can most easily write on.

Two laundering channels are closed alongside the venue text: the lens **mechanism** (the previous call's own sentence about the same untrusted criteria, persisted in `lens_verdicts` and replayed into VERIFY/GROUND) and the agent_trader **thesis** (the arm's own output, persisted and replayed as memory).

`agent_trader`'s candidate block had a live asymmetry: `description` got `.replace("\n", " ")` and `question` got nothing, so a newline inside a question forged a candidate line — a market the arm was never offered, complete with an id and a price. Both now take the full scrubber, which collapses every whitespace class rather than just `\n`.

## Bounds — what a model can and cannot see

Every pre-existing slice is **preserved and scrubbed in place** (`[:200]`, `[:300]`, `[:1000]`, `[:1500]`, `[:2000]`, `[:40000]`). No site sees less content than before; scrubbing only removes control/invisible characters and collapses runs of whitespace.

`resolution_lens`'s ~4500-char HEAD+TAIL criteria cap is the one that needed care, and it is unchanged in shape: the scrubber runs over the **whole** description first and the cap is applied to the result, so the TAIL — where the decisive qualifier usually lives — can never be truncated by normalization. A test pins the head/tail split and that tail content survives, and a second pins `criteria_char_cap == 4500`.

**Disclosed behavior changes, all small but real:**
- Angle brackets escape to `\u003c` / `\u003e`, so a criteria text containing many `<` or `>` spends more of its cap on escapes and loses slightly more of its trimmed MIDDLE (never its head or tail).
- New ceilings land only on values that previously had none: questions at 1000 (the `format_market_context` bound; no venue question approaches it), ids 120, category 60, thesis 400, replayed memory/open-book questions 300. A thesis or question longer than those is truncated where it previously was not.
- Market ids are scrubbed before display. Every consumer validates returned ids against the candidate set, so an id the scrubber would alter simply fails to match and no trade is taken — fail-closed, and the same treatment `format_market_list` already gives ids since #403.

## Not covered

- **#405 item 1** — the deterministic token-overlap + resolution-date post-check on LLM-proposed pairs. Untouched; that is the structural fix and prompt hardening does not substitute for it.
- `cross_venue_arb.py` and `entailment_arb.py` — owned by a parallel branch.
- `{world_model}` in `agent_analyzer` / `strategic` still sits outside a boundary. It is LLM-authored and persisted, i.e. the same laundering shape as thesis/mechanism, but it is the agent's sanctioned state rather than venue text; flagged in the #410 review, still open.
- Lower-severity sites from that review are unchanged: `news_triage.py`, `evidence_distiller.py`, `correlation.py`.

## Verification

CI-identical container run:

```
45 files skipped due to complete coverage.
Required test coverage of 65.0% reached. Total coverage: 66.54%
2188 passed, 5 skipped, 1 deselected in 215.81s (0:03:35)
All checks passed!
```

New tests in `tests/test_prompt_hardening_405.py`: per-codepoint scrubbing, the invisible-instruction and zero-width-in-delimiter attacks, the ZWJ decision, and per-site coverage driving the **real** call paths (`_verdict`, `_verify_mechanism`, `_ground`, `_read_family`, `_extract`, `_verify`, `_audit_filing`, `deep_research`) with a hostile payload that tries to close the boundary, issue instructions, and forge a line of structure — asserting exactly one delimiter pair per prompt and that the payload survives only as quoted data.

### DIFF
--- a/auramaur/nlp/prompts.py
+++ b/auramaur/nlp/prompts.py
@@ -7,7 +7,36 @@ import re
 import unicodedata
 from urllib.parse import urlparse
 
-_CONTROL_OR_BIDI = re.compile(r"[\x00-\x08\x0b\x0c\x0e-\x1f\x7f\u202a-\u202e\u2066-\u2069]")
+# Characters that render as nothing (or as direction changes) and therefore
+# carry text a human reviewer never sees. Three families, one class:
+#   * C0/C1 controls and DEL: \x00-\x08\x0b\x0c\x0e-\x1f\x7f
+#   * BiDi overrides/embeddings/isolates and the LRM/RLM marks: U+202A-202E,
+#     U+2066-2069, U+200E-200F
+#   * zero-width and invisible formatting: SOFT HYPHEN U+00AD, ZWSP/ZWNJ/ZWJ
+#     U+200B-200D, WORD JOINER + invisible operators U+2060-2064, BOM /
+#     ZWNBSP U+FEFF, and the Unicode TAG block U+E0000-E007F (a full
+#     invisible ASCII channel).
+#
+# 2026-08-06 (#405 item 2): the last family was NOT stripped. Verified
+# empirically against this exact pipeline: NFKC leaves every one of them in
+# place and none has the White_Space property, so `" ".join(text.split())`
+# keeps them too. An instruction spelled in tag characters, or a
+# `</UNTRUSTED_...>`-shaped payload with ZWSPs wedged between the letters,
+# survived into the prompt looking like nothing at all.
+#
+# U+200D ZERO WIDTH JOINER is load-bearing in legitimate emoji sequences
+# (family/profession/flag ZWJ sequences collapse to their parts without it).
+# Stripped here anyway, deliberately: these prompts carry market questions,
+# resolution criteria, news text and SEC filings to a model that reasons about
+# them analytically. Emoji glyph fidelity has no bearing on that judgment,
+# while an invisible joiner inside a delimiter-shaped payload does. If a prompt
+# ever needs emoji rendered faithfully, it needs its own formatter, not a hole
+# in this one.
+_CONTROL_OR_BIDI = re.compile(
+    "[\x00-\x08\x0b\x0c\x0e-\x1f\x7f"
+    "\u00ad\u200b-\u200f\u202a-\u202e\u2060-\u2064\u2066-\u2069\ufeff"
+    "\U000e0000-\U000e007f]"
+)
 
 
 def format_untrusted_text(value: object, limit: int) -> str:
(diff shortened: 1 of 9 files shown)


DarriEy:
Review (Claude, 2026-08-06): **MERGE-READY.** The three disclosed truncation risks are the right things to have disclosed, and I measured all three against the live DB (89,505 markets, 395 stored theses) — data the authoring agent had no access to:

| Risk | Measured | Verdict |
|---|---|---|
| Market id altered by scrub -> id-validation fails -> no trade (fail-closed) | **0 of 89,505** ids alter under NFKC+collapse; **0** exceed the 120 bound | cannot fire on real data |
| Thesis 400-char ceiling | 395 theses, median 214, max 500, **2 over 400** (0%) | negligible; loses <=100 tail chars on 2 rows |
| Question 1000-char ceiling | max 1410, **29 of 89,505 over** (0.03%) | negligible |

So no live behavior change from the new ceilings, and the fail-closed id path is unreachable in practice — good, because that one would have been silent.

**On the substance:** widening `_CONTROL_OR_BIDI` past the exact codepoints #405 listed was the right call — they are one class, and stopping at the enumerated ones would have left U+00AD and U+2060 as arbitrary survivors. Finding that **LRM/RLM (U+200E-200F) also survived** is a genuine addition: that is a hole in the BiDi defense itself, not the zero-width family, and neither the issue nor my #410 review caught it.

The ZWJ decision is correctly reasoned and correctly pinned by a test that asserts the deliberateness. Emoji fidelity buys nothing in a prompt carrying resolution criteria to a model reasoning analytically, and a prompt that ever needs it should get its own formatter rather than a hole in the shared one.

**The resolution_lens handling is the part I would have got wrong.** Scrubbing the whole description and THEN capping — rather than scrubbing capped segments — is right for a reason worth recording: NFKC can expand, so scrubbing after the cap would let expansion push content past the boundary and eat the tail, which is the half the 4500 head+tail cap exists to protect. Both the shape and the cap value are pinned by tests.

Promoting `strategic._scrub` to `prompts.format_untrusted_block()` is the correct structural move — six pillars sharing one definition beats six importing a private symbol across packages.

Container verification reproduced as stated: 2188 passed, coverage 66.54% >= 65.0, ruff clean.

Closes #405 item 2. Item 1 (deterministic post-check on LLM-proposed pairs) and the two arb-pillar prompt sites are owned by a parallel branch.

PR #416, opened 2026-08-06
Give the ExecutionGateway a cancel contract, deliberately ungated

DarriEy:
## The problem

CLAUDE.md rule 3 makes `ExecutionGateway` the single mechanical path for an order. But only **placement** ever had a contract. There was no cancel method on the gateway at all, so every cancel in the codebase reached past the choke point straight to `ExchangeClient.cancel_order`:

| Call site | Status before |
|---|---|
| `strategy/market_maker.py` ×3 (`_place_two_sided`, `_cancel_quote` ×2) | **Direct rule violations** — a pillar calling a raw exchange order method |
| `broker/execution_gateway.py:476` (paired leg-A unwind) | The enforcer violating its own rule |
| `bot.py` / `bot_order_monitor.py` / `bot_arb_execute.py` | Orchestration, outside the pillar perimeter |

`scripts/sweep_invariants.py` (#404) even carried a comment excusing the gap: `cancel_order` was deliberately excluded from the `strategy-raw-order` dimension because "a strategy has nowhere else to route." That was true, and it made the rule unenforceable rather than false-negative-prone. The rule asserted an invariant its own enforcer violated.

## The design decision: routing and audit, NOT a gate

`ExecutionGateway.cancel_resting()` **cannot refuse.**

- **No risk checks.** None of `submit()`'s fifteen apply. A placement adds exposure and is gated accordingly; a cancel *removes* exposure, so those checks are not merely unnecessary here — they are backwards. There is no code path that returns "blocked".
- **No kill-switch check, and a dated comment saying so** at the exact point where a reader would otherwise "helpfully" add one.

  This is the critical inversion to avoid. **#408** (merged 2026-08-06, `1702b43`) made arming the kill switch *retire live exposure* by cancelling resting orders. A cancel that the kill switch could block would therefore trap the operator inside the exposure the emergency stop exists to shed — the emergency stop would become an exposure lock. The kill switch belongs on `place_order`, where it already is, and nowhere on this path.

What the contract *does* provide:

- **Paper interception** — a paper order never reaches a venue client. This is a real boundary, not a formality: quoting runs paper orders through the **live client object** (paper-ness is per-order `dry_run`, not a separate adapter). Interception is an OR of an explicit `is_paper` flag and the `PAPER-` id prefix, so a caller that computes the flag wrongly still cannot leak a paper id to the venue.
- **Terminal-state bookkeeping** — the `trades` mirror goes `'cancelled'` instead of lingering `'pending'` forever. Same repair `_cancel_resting_live_orders` and the order-monitor TTL sweep each hand-rolled (#94), scoped to `status = 'pending'` so a mis-supplied id can never rewrite a settled row.
- **A uniform `CancelResult`** (`cancelled` / `paper` / `rejected` / `failed` / `skipped`, with `.ok`) and **one greppable audit event** (`gateway.cancel`).
- **Exception containment** — a venue error is *reported*, never raised. These run in a quoting loop, and callers already treated cancels as best-effort.

### Signature

```python
async def cancel_resting(
    self, order_id: str, *,
    exchange: ExchangeClient | None = None,
    exchange_name: str | None = None,
    is_paper: bool = False,
    source: str = "",
    reason: str = "",
) -> CancelResult
```

Derived from the four real call sites. `exchange`/`exchange_name` default to the gateway's own exactly as `submit()` resolves them, and a cross-venue caller passes its own the way `submit_exit` and `submit_paired` already do — no parallel mechanism. `is_paper` carries `OrderResult.is_paper` where the caller has it. `source`/`reason` preserve the attribution and the per-site specificity of the four distinct log event names being replaced.

Named `cancel_resting`, not `cancel`: `cancel` collides with `task.cancel()` in the AST-based registry perimeter, and not `cancel_order` so that the raw venue verb stays uniquely greppable and statically bannable inside pillars.

## Migrated

- **`market_maker.py` ×3** → `cancel_resting`. The pillar no longer names `cancel_order` anywhere, and the `PAPER`-prefix tests it used to make are gone: interception is the choke point's job, and duplicating it in the pillar is how the two drift apart. Loop behavior is otherwise preserved — same venue round-trip, same best-effort, exceptions still never reach the loop.
- **`execution_gateway.py:476` (paired leg-A unwind)** → `cancel_resting`. Semantics unchanged (same `pending and not is_paper` guard, same best-effort unwind, exceptions contained); the unwind additionally drives leg A's trades row terminal instead of leaving it `'pending'`.

## Unmigrated by choice

The three bot-level sites — `bot.py:_cancel_resting_live_orders` (shutdown/kill-switch sweep), `bot_order_monitor.py` (TTL reaper), `bot_arb_execute.py` (cross-exchange rollback) — are orchestration, not pillars, and stay direct in this PR.

My view is that they *should* eventually converge on this contract, for a reason independent of the rule: two of the three hand-roll the exact `UPDATE trades SET status='cancelled'` this method now owns, and the third (arb rollback) does not, so a rolled-back leg's row stays `pending`. That is the same triplicated-tail duplication the gateway was extracted to delete. The blocker is not the rule — it is that each has orchestration the contract does not yet carry: the shutdown sweep iterates `client._live_pending` and prints an operator summary, the TTL reaper must keep an unacknowledged order *tracked* rather than dropping it, and the rollback escalates a failed cancel to a critical alert. Those are separable, but folding them in blind would change kill-switch and reconciliation behavior on the same commit that establishes the contract, so they belong in a follow-up with their own tests.

## Re-armed enforcement

- `sweep_invariants.py`: `cancel_order` added back to `strategy-raw-order`; the excusing comment is replaced by a dated note on why it was excluded and what fixed it. New unit test proves the dimension has teeth (raw cancel in a pillar → finding; routed cancel → none; the gateway itself → none).
- `tests/test_strategy_protocol.py`: `_PLACEMENT_METHODS` → `_RAW_ORDER_METHODS`, now including `cancel_order`, so a pillar in `NO_DIRECT_PLACE_MODES` fails statically on a raw cancel.
- `exposure_registry.py`: market_maker's cancels re-pinned to `cancel_resting`; both verbs in `SENSITIVE_METHODS`, so the single remaining raw `cancel_order` on this side of the perimeter is the one *inside* the contract, registered against `gateway_boundary`.
- `.sweep-baseline.json` unchanged — this migration cleared no baselined findings (`cancel_order` was never a baselined dimension).

## CLAUDE.md

Rule 3 gains the cancel contract **and** an explicit statement that the path is deliberately ungated, with the reason. The documentation half of the fix: the rule is now true.

## Tests

`tests/test_execution_gateway.py`, house conventions:

- (a) a pillar cancel routes through the gateway and reaches the exchange client
- (b) a paper order's cancel never reaches a live client (id shape, explicit flag, and via the pillar)
- (c) **a cancel succeeds while the kill switch is armed** — the regression that matters most, checked behaviorally *and* structurally (`cancel_resting`'s body, docstring and comments stripped, may not mention the kill switch)
- (d) a venue exception is contained and reported, not raised; a declined cancel is `rejected`, not silently successful
- (e) the trades row reaches a terminal state — and a `filled` row is never rewritten, and a declined cancel leaves the row `pending`
- plus: the paired unwind goes through the contract and drives leg A terminal

## Verification

CI-identical container: **2166 passed, 5 skipped**, coverage **66.35%** (≥ 65.0), ruff clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,6 +13,19 @@
      order method.
    - Prediction-market exits use the gateway exit contract. The sole
      off-gateway exception is the declared IBKR equity path.
+   - **Cancels** use `ExecutionGateway.cancel_resting()`. It covers the other
+     end of an order's lifecycle: strategy pillars must not call an exchange's
+     raw `cancel_order` any more than its raw `place_order`. Bot-level
+     orchestration (shutdown sweep, order-monitor TTL, arb rollback) is
+     outside the pillar perimeter and still cancels directly.
+   - **The cancel path is deliberately UNGATED, and must stay that way.** It
+     carries none of `submit()`'s risk checks and — the part that matters —
+     **no kill-switch check**. A cancel reduces exposure, and arming the kill
+     switch RETIRES live exposure by cancelling resting orders (#408); a
+     blockable cancel would therefore trap the operator inside the exposure the
+     emergency stop exists to shed. `cancel_resting` provides routing, paper
+     interception, terminal-state bookkeeping, audit logging and exception
+     containment. It can never return "blocked".
 4. **Never hardcode API keys.** All secrets come from environment variables.
 5. **Never read `.env` files.** They contain secrets. Use `.env.example` for reference.
 6. **Never force-push to main.**
@@ -21,9 +34,10 @@
 When making git commits, use `Assisted-by: Claude (Anthropic)` in the commit message body instead of `Co-authored-by`. The human author should always be the sole git author of record.
 
 ## Architecture
-- Prediction-market placements flow through `auramaur/broker/execution_gateway.py`;
-  its methods delegate to the appropriate exchange adapter. Strategy pillars
-  must not call raw exchange order methods.
+- Prediction-market placements AND cancels flow through
+  `auramaur/broker/execution_gateway.py`; its methods delegate to the
+  appropriate exchange adapter. Strategy pillars must not call raw exchange
+  order methods — neither `place_order` nor `cancel_order`.
 - Paper trading interception happens in `auramaur/exchange/paper.py`.
 - `auramaur/risk/manager.py` is the approval authority for directional entries.
   Structural strategies use their declared, test-enforced gateway contracts;
--- a/auramaur/broker/execution_gateway.py
+++ b/auramaur/broker/execution_gateway.py
@@ -17,6 +17,11 @@ evaluation stays with the caller — the caller passes the already risk-approved
 pair (the arb pillars): it builds BOTH orders before placing EITHER, places A
 then B, and unwinds a live-pending A if B fails — preserving the atomicity a
 per-leg ``submit`` could not.
+
+``cancel_resting`` closes the other half of a resting order's lifecycle. It is
+a ROUTING AND AUDIT contract, not a gate: unlike a placement, a cancel reduces
+exposure, so it carries none of ``submit``'s risk checks and cannot refuse.
+See its docstring for why the kill switch must never reach it.
 """
 
 from __future__ import annotations
@@ -78,6 +83,29 @@ class ExecutionResult:
     result: OrderResult | None = None
     fill: Fill | None = None
     reason: str = ""
+
+
+@dataclass
+class CancelResult:
+    """The outcome of a gateway cancel. Never an exception, never a refusal.
+
+    ``status`` is one of:
+
+    ``cancelled``  the venue acknowledged the cancel.
+    ``paper``      a paper order — intercepted, never sent to a venue client.
+    ``rejected``   the venue declined (returns False, e.g. already matched).
+    ``failed``     the venue call raised; the exception was contained here.
+    ``skipped``    no usable order id was supplied — nothing to cancel.
+    """
+
+    order_id: str
+    status: Literal["cancelled", "paper", "rejected", "failed", "skipped"]
+    reason: str = ""
+
+    @property
+    def ok(self) -> bool:
+        """True when the order is no longer resting because of this call."""
+        return self.status in ("cancelled", "paper")
 
 
 class ExecutionGateway:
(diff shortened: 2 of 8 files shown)


DarriEy:
Review (Claude, 2026-08-06): **MERGE-READY.** The design brief said ungated-by-construction and the implementation honors it; three judgment calls improved on what I specified.

**Naming was the right catch.** `cancel_resting` rather than `cancel` because `cancel` collides with `task.cancel()` in web/broker.py and SENSITIVE_METHODS matches by attribute name — adding `cancel` there would have broken the web-perimeter test. And not `cancel_order`, so the raw venue verb stays uniquely bannable inside pillars. I would not have seen either.

**Paper-ness as an OR of the flag and the `PAPER-` id prefix** is better than what I asked for: a caller that computes the flag wrongly still cannot leak a paper id to the venue. That is the right shape for an interception boundary — belt and braces, at the choke point rather than duplicated in each pillar.

**The kill-switch requirement is met and tested**, which was the one non-negotiable: a cancel must succeed while the switch is armed, because #408 made the switch retire exposure *by cancelling*. A blockable cancel would have inverted that entirely.

**On the three market_maker deltas — all disclosed, all acceptable, one worth watching.** The added serialized UPDATE after each successful cancel sits inside the `op_timeout` budget on a quoting loop. It is a single indexed statement, after the venue round-trip, in a try/except so a busy DB cannot surface as a cancel failure, on a path handling a handful of orders per refresh. Fine — but it is the one thing in this PR that touches latency on a live loop, so if maker behavior looks off after deploy, look here first. Failed cancels moving from debug to warning is strictly more signal. Relocating the PAPER-prefix checks from the pillar to the gateway is the point of the exercise, not a side effect: duplicated interception is how two implementations drift.

**Your answer on the bot-level sites is better reasoned than my scoping.** You are right that the case for migrating them is duplication rather than the rule — two hand-roll the same terminal-state UPDATE the contract now owns, and `bot_arb_execute.py`'s rollback does not, so a rolled-back leg's row stays `pending` indefinitely. That is a real bug my scoping would have left buried. Equally right that folding them in now would change kill-switch and reconciliation behavior on the same commit that establishes the contract. Follow-up with its own tests, and I agree `bot.py:_cancel_resting_live_orders` is the site to leave alone longest — it is the kill switch's own exposure-retirement path.

**Useful independent finding:** `.sweep-baseline.json` is already stale on origin/main by two entries (gap_audit untrusted-repr x2, fixed by #410 without a baseline refresh). Harmless today since a superset baseline only over-accepts and the sweep still exits 0 — but it is the first live instance of the no-stale-detection limitation flagged on #404, and it argues against wiring that sweep as a blocking CI gate until stale detection exists.

Container verification: 2166 passed, coverage 66.35% >= 65.0, ruff clean.

CLAUDE.md rule 3 now describes a world that exists.

PR #426, opened 2026-08-07
Require explicit live authority grants

DarriEy:
## Summary
- replace directional graduation exemptions with explicit grants scoped by strategy, venue, and category
- enforce grant stake caps, review deadlines, settlement thresholds, and cumulative-loss stops
- fail closed to paper when grants expire, breach limits, or evidence cannot be read
- retain a canonical allowlist for structural arbitrage and market-making exemptions
- expose venue scope in the graduation report and document the configuration shape

## Verification
- Ruff checks passed
- focused tests: 82 passed
- full suite: 2331 passed, 6 skipped

## Stack
Depends on #425. Retarget this PR to main after #425 merges.

### DIFF
--- a/auramaur/bot.py
+++ b/auramaur/bot.py
@@ -6,7 +6,7 @@ import asyncio
 import logging
 import time
 from pathlib import Path
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from auramaur.killswitch import kill_switch_present
 from typing import TYPE_CHECKING
 
@@ -33,6 +33,7 @@ from auramaur.nlp.cache import NLPCache
 from auramaur.nlp.calibration import CalibrationTracker
 from auramaur.risk.manager import RiskManager
 from auramaur.risk.portfolio import PortfolioTracker
+from auramaur.risk.exit_lifecycle import ExitState, record_exit_state
 from auramaur.strategy.engine import TradingEngine
 from auramaur.strategy.market_maker import MarketMaker
 from auramaur.strategy.news_reactor import NewsReactor
@@ -508,6 +509,11 @@ class AuramaurBot(
                             retry_at = self._exit_failures.get(exit_key)
                             if retry_at is not None and time.monotonic() < retry_at:
                                 continue
+                            await record_exit_state(
+                                self._components.db, pos, ExitState.REQUESTED,
+                                exchange=name, reason=reason.value,
+                                increment_attempt=True,
+                            )
                             log.info(
                                 "exit.triggered",
                                 exchange=name,
@@ -530,9 +536,26 @@ class AuramaurBot(
                             if ok:
                                 self._exit_pending.add(exit_key)
                                 self._exit_failures.pop(exit_key, None)
+                                await record_exit_state(
+                                    self._components.db, pos, ExitState.ORDER_WORKING,
+                                    exchange=name, reason=reason.value,
+                                )
                             else:
                                 self._exit_failures[exit_key] = (
                                     time.monotonic() + _EXIT_RETRY_BACKOFF_SECONDS)
+                                retry_wall = datetime.now(timezone.utc) + timedelta(
+                                    seconds=_EXIT_RETRY_BACKOFF_SECONDS)
+                                dust = (
+                                    name == "polymarket" and pos.size < 5
+                                ) or (
+                                    name == "kalshi" and pos.size < 1
+                                )
+                                await record_exit_state(
+                                    self._components.db, pos,
+                                    ExitState.UNSALEABLE_DUST if dust else ExitState.RETRYABLE,
+                                    exchange=name, reason=reason.value,
+                                    next_retry_at=None if dust else retry_wall,
+                                )
                                 log.warning(
                                     "exit.suppressed_until_retry", exchange=name,
                                     market_id=pos.market_id, reason=reason.value,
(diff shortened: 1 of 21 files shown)


DarriEy (commented):

Eleven finder angles plus adversarial verification against the rebased head (`749dff5`). Overall: this is the right design — explicit, scoped, expiring grants with fail-closed defaults are exactly the answer to the exemption/force-paper-latch incident class. But several of the **bounds the grant promises are not real yet** (three fail-open holes and one cap-defeat), the deploy itself is a silent liveness cliff, and the arb executor sits outside the whole scheme. The enforcement floor needs to be actual before this governs money.

### Deployment-blocking

1. **Deploying halts all live directional trading, silently** — `graduation.py:99`. Tracked config ships `live_authority: {}` while directional `exempt_strategies` are revoked; the operator promotions in the production override stop working with no migration path and no alarm beyond per-call log noise. Grants must be hand-written into the gitignored override *before* deploy, and nothing checks or reports that. Needs a startup cross-check (unmatched exemptions/grants → loud, once) and a documented cutover.

2. **The documented grant example doesn't validate** — `config/defaults.yaml:231`. The template omits the required `review_after_settlements` field, so copying it into the override crash-loops the container at Settings() load. The one workflow this feature ships for fails on its own documentation.

### Fail-open holes in the promised bounds

3. **The loss stop is blind to label drift** — `graduation.py:157`. Evidence matches exact venue/category strings stamped at *realization* time from the markets row; once the markets row ages out, settlements land blank-labeled and the `WHERE venue=? AND category=?` never sees the loss — `stop_loss_usd` fails open exactly on long-dated positions.

4. **The review counter never counts exits** — `graduation.py:153`. Only `kind='settlement'` rows count, but live exits realize P&L as `kind='sell'` (the dominant pattern; `ledger_report._REALIZATION_KINDS` includes both). A strategy that exits before resolution accrues realized P&L while `review_after_settlements` stays at zero — the pre-registered review boundary is inert.

5. **The stop is realized-only; concurrent exposure is unbounded** — `graduation.py:166`. `max_stake_usd` is per-order and `stop_loss_usd` sums realized rows, so a "bounded" grant can hold dozens of concurrent positions with aggregate/unrealized exposure bounded by nothing grant-specific until settlements land — months, for long-dated markets. At minimum the grant needs an open-position or committed-notional cap.

6. **The stake cap is breakable by the venue minimum bump** — `manager.py:491`. The cap is a post-hoc `min()`, but `prepare_order` bumps sub-minimum orders UP to the 5-token/$1 floor (Kalshi: 1 contract) — a small cap is silently exceeded at the venue. The adjacent zero-multiplier guard documents this exact defeat vector and hard-rejects; the grant cap should do the same when the bump would cross it.

### Mode and path interactions

7. **`off` drops the caps; `observe` enforces them** — `graduation.py:93`. The grant layer is nested inside the mode switch backwards on both sides: `mode=off` returns `_LIVE_FULL` before the grant check (stake cap, stop, expiry all vanish for granted strategies), while `mode=observe` runs grants before the observe neutering, changing behavior in the mode contractually pinned to never change behavior. The existing contract test can't see it because it never exercises a grant.

8. **The arb executor places live orders after paper-grade checks** — `bot_arb_execute.py:117` (pre-existing, widened by this PR). All three arb paths evaluate under the default `strategy_source="llm"` and discard `decision.force_paper`; the live-only containment checks are skipped for what risk believes is a paper entry, yet the legs are placed with `dry_run=not settings.is_live`. Every new fail-closed grant state this PR adds routes more entries into that gap.

9. **Venue-less `decide()` calls silently skip grants** — `resolution_lens.py:611`. The `venue=""` default means omitting the argument opts out of grant matching with no signal; the lens's live-category prioritization already does, so two components can disagree about one cell's mode.

### Timing, observability, plumbing

10. **Date-anchored window attributes pre-grant P&L** — `graduation.py:159`. `granted_at` is a midnight date string, so same-day pre-grant losses and settlements of positions entered before the grant count against the stop — a grant can be born tripped. Anchor on a timestamp and (ideally) on entries the grant actually authorized.

11. **Inert grants produce zero signal** — `graduation.py:131`. Matching is exact on the un-normalized axes (strategy key, category); a typo'd grant validates clean and simply never matches — no startup cross-check, no report row, no log. A tripped grant logs; an inert one never does. Same class as the deploy cliff: absence-of-liveness is this repo's least-visible failure mode.

12. **Grant path is uncached and flappy** — `graduation.py:96`. Runs the pnl_ledger aggregate on every evaluate (the ladder path has a 300s TTL cache), and a transient locked-DB error silently flips the cell to paper for a tick via `grant_evidence_unavailable`, then back — live/paper oscillation recorded only by an exception log.

13. **Per-call ERROR flood for stale exemptions** — `graduation.py:102`. The `directional_exemption_ignored` branch logs at ERROR on every decide() for the affected strategy — after deploy, with the override unchanged, that's continuous. Reject at config validation or warn once per cell.

14. **Interim-manager counting breaks under the venue split** — `interim_manager.py:113`. The per-venue GROUP BY double-counts one (strategy, category) cell toward `sunset_after_live_cells`, and `operator_grant` status isn't in `_GRADUATED` — granted cells neither count toward sunset nor claim category ownership, so the manager can stand down early *and* keep proposing into granted categories.

15. **The grant SQL never runs in CI** — `tests/test_live_authority.py:36`. Every test mocks `db.fetchone`, and production wraps the query in a blanket except that fails closed — so a future schema change kills the entire feature silently (permanent `grant_evidence_unavailable` paper-forcing) with green tests. Use the real-DB harness pattern from `test_graduation.py` for at least one grant-evidence path.

PR #427, opened 2026-08-08
Persist execution decision lineage

DarriEy:
## Summary
- persist decision IDs through gateway trades and immediate, deferred, and monitored fills
- add schema v52 with safe legacy backfill and lineage indexes
- audit missing, orphaned, and mismatched decision, trade, fill, and realized-PnL links
- surface execution-lineage health in auramaur doctor while allowing legitimate manual and external fills

## Verification
- Ruff checks passed
- focused lineage and execution tests: 75 passed
- full suite: 2334 passed, 6 skipped

## Stack
Depends on #426. Retarget this PR after #426 merges.


DarriEy (commented):
## Multi-agent review — 11 findings, not mergeable as-is

Ten independent finder angles over the PR head; each finding below was verified against the code (the top ones independently by 3–5 angles). Overall: the goal is right and the migration/conventions hygiene is clean, but as built the audit **manufactures its own violations** and misses the largest real lineage-losing path. All fixes are tractable.

**Stack note:** this branch is based on the pre-review-fix `#425` head (`8c0b38e`). Rebase onto current `feat/durable-exit-reconciliation` (`9216d84`) first — expect a conflict in `tests/test_exit_lifecycle.py` (both branches edited the schema-version assert) and take care in `bot_order_monitor.py` (the fix commit added `_record_exit_terminal` calls beside the lines this PR touches).

### Critical

1. **Restarts drop lineage, and the audit then reports normal operation as corruption** — `bot_order_monitor.py:186`. `decision_id` lives only on the in-memory Order; both venues' `reconcile_open_orders` rebuild adopted resting orders without it (`client.py:662`, `kalshi.py:874` — client.py already fetches the trades row but copies only `strategy_source`). Post-restart fills book `decision_id=NULL` against a trades row that has the id: on Kalshi `fill_trade_lineage_mismatch` then fires forever after every routine redeploy-with-resting-orders; on Polymarket the adopted order's condition-hash `market_id` evades the check's join, so the main leak is silently invisible to the audit built to catch it. Fix: read `decision_id` back in the same trades lookup reconcile already does, or derive it in `record_fill` when the in-memory value is None.

2. **`governed_trade_missing_decision` is born-red on months of history** — `data_quality.py:64`. Trades data predates `decision_snapshots` (created 2026-07-19, #315) by ~4 months; those rows can never be backfilled and count as violations forever. `data_quality.py:33-36` documents the file's own rule for exactly this ("earlier rows predate the contract and would otherwise count as violations forever") — the new checks skip it. A permanently-warning doctor line buries every future real break. Fix: watermark to post-deploy rows (e.g. anchored on `MIN(decision_snapshots.observed_at)`) and leave history honestly NULL.

3. **The v52 backfill mints violations its own checks then flag** — `db/database.py:410`. Matching on `(signal_id, strategy_source)` alone: a signal submitted paper then re-submitted live shares ONE snapshot (`INSERT OR IGNORE` under the UNIQUE constraint), pre-v43 snapshots carry `is_paper` defaulted to 1, NULL-`strategy_source` trades never match (SQL NULL equality — asymmetric with the check's own COALESCE), and nothing requires the snapshot to predate the trade it explains. Immutable wrong lineage, audited as corruption from migration day one. Fix: leave pre-migration rows NULL; if any backfill is kept, require `d.observed_at <= trades.timestamp` and match `is_paper`.

### High

4. **The concurrent-arb path can never carry lineage, invisibly** — `execution_gateway.py:921`. `record_external_fill` omits the kwarg and `_record_result` never reads `order.decision_id`; it also forces `signal_id=None`, so no contract check can ever flag the gap. Same silent-gap class as the paired-arb snapshot bug documented at `execution_gateway.py:629-639`.

5. **Three carriers for one value; omitting the kwarg erases the field** — `execution_gateway.py:882`. Caller stamps at `:312`/`:647`, re-stamp at `:882`, plus a parallel kwarg — and the unconditional `order.decision_id = decision_id` resets a pre-set field to None on any path that omits the kwarg (submit_exit today). That silently kills the deferred-fill and `DecisionTracker.mark_fill` reads. Fix: make `order.decision_id` the sole carrier and delete the kwarg.

6. **NULL-signal capture race becomes durable wrong lineage** — `research/polymarket_strategies.py:244`. `capture()`'s select-back (`WHERE signal_id IS NULL AND strategy_source=? ORDER BY id DESC`) can return a concurrent submission's snapshot; this PR promotes that race's output into immutable trades/fills rows that `trade_decision_mismatch` then reports. Fix: use the insert's `lastrowid`.

7. **`orphan_ledger_fill` flags states the codebase documents as legitimate** — `data_quality.py:101`. Repair waves deliberately deleted/renamed fills rows while keeping ledger rows (`manual_trades.py:27-29` guards this exact state), and the planned phantom-fill purge would trip it too. No epoch bound, no allowlist, no remediation path.

### Medium

8. **O(fills×trades) audit; the index it needs doesn't exist, the ones it adds have no readers** — `data_quality.py:89`. Two correlated EXISTS probes per fills row keyed on `order_id`, but trades has no `order_id` index (probes walk `idx_trades_market` — worst on the busiest markets); the v52 backfill pays the same join at startup on the live file. Meanwhile `idx_trades_decision`/`idx_fills_decision` have zero readers. Fix: add `idx_trades_order`, drop the reader-less pair until a read path exists.

9. **Doctor classification by hand-maintained prefix tuple; full audit run discarded** — `diagnostics.py:294`. The 8-prefix tuple mirrors names declared in another module (drift silently drops checks from the operator's alarm surface), the doctor runs all 16 audit queries and throws away the 8 dataset results (including the heaviest `json_each` join in the file), and the bare `except Exception` collapses disk corruption and a missing table into one string. Fix: a `category` field on the check declarations; request only the execution subset; log the exception.

10. **The v52 migration has no direct test** — `tests/test_exit_lifecycle.py:62`. The fresh-schema assert is tautological against `SCHEMA_VERSION`, and nothing calls `_migrate_v51_to_v52` directly — the ALTER + heuristic backfill first execute against the production DB. Replicate the DROP TABLE + direct-call pattern (`test_exit_policy.py:221`, and the v51 test on the rebased base).

### Design question

11. **`decision_id` denormalized into both trades and fills** — `db/models.py:327`. Every non-NULL fill value is written by the same call (or derivable via the exact join the migration itself uses) as its trades row's; `fill_trade_lineage_mismatch` exists solely to police drift between the two copies. A single authoritative column deletes a column, a model field, two threading sites, half the backfill, an index, and three contract checks.

PR #428, opened 2026-08-08
Calibrate earlier profit taking on holdout

DarriEy:
## Summary
- sample HOLD counterfactuals hourly and retain 30 days with fewer rows than the prior three-day minute cadence
- replace the double-counting exit audit with replay over completed observed episodes
- compare earlier-banking candidates against deployed policy using paired dollar outcomes
- select on the oldest 70 percent and score the winner once on the newest 30 percent
- require a positive paired 95 percent lower bound in both periods and never mutate production configuration
- exclude open, right-censored, and repeated same-position episodes from independent evidence

## Verification
- Ruff checks passed
- focused exit-policy tests: 77 passed
- full suite: 2343 passed, 6 skipped
- read-only command against current database correctly returned no recommendation because no completed target or trailing episodes survive retention

## Stack
Depends on #427. Retarget this PR after #427 merges.


DarriEy (commented):
## Multi-agent review — 12 findings, not mergeable as-is

Ten finder angles plus an adversarial verification pass (several candidate findings were refuted and dropped; the ones below survived). Overall: the train/holdout split and paired-LCB gate look rigorous, but the **sample construction undermines the question being asked** — the statistics are careful arithmetic over a biased population. The loss-path censoring (finding 2) needs a design answer before the rest is worth polishing.

**Stack note:** based on the pre-review-fix `#425` via the stack; rebase onto current branch heads before iterating — `risk/portfolio.py` was touched by the #425 fix commit (`9216d84`).

### Critical — sample construction

1. **Episodes stitch two different real positions together** — `risk/exit_calibration.py:92`. Only PROFIT_TARGET/TRAILING_STOP write terminal rows; stop-loss, edge-erosion, time-decay, dust, settlement, and manual closes write nothing. A later position re-entering the same (exchange, market, token, book) key inherits the dead position's HOLD rows: e.g. position A holds at +35% then stop-losses (no row); position B re-enters and hits PROFIT_TARGET — `bank_earlier` harvests A's +35% (A actually lost) against B's unrelated terminal, one-directionally inflating deltas. The keep-first dedup at `:152` then retains exactly the contaminated episode. Can fabricate a false-positive recommendation.

2. **Winners-only sample censors the benefit being measured** — `risk/exit_calibration.py:10`. Because loss-path exits write no terminal rows, episodes that peak and then ride down into the stop never enter the sample — yet those are precisely where earlier banking wins. On recorded-win episodes, earlier banking mostly forfeits upside, so the tool systematically answers "no candidate improves deployed policy" regardless of the true full-book delta. The docstring attributes exclusion to right-censoring only, hiding this second, systematic censoring. Design fix: loss exits must write terminal rows (which also fixes finding 1).

3. **Fee currencies mixed inside single paired deltas** — `risk/portfolio.py:549`. TRAILING_STOP rows pass `pnl_pct` twice, so `net_pnl_pct` = gross with fees 0.0; replay then pairs fee-net HOLD marks against fee-gross trailing terminals. Every trailing-terminated pair understates the candidate by the uncharged exit fee, penalizing the earliest-firing candidates most; the doc claim "the estimate uses the fee-net decision mark" (docs/take-profit-policy.md:76) is false for trailing rows.

### Critical — epoch discipline

4. **The "deployed" baseline replays today's config over historical rows** — `scripts/calibrate_exit_policy.py:69`. Rows persist no record of the trailing config in force when written; targets replay per-row (epoch-faithful) but trailing replays globally-current, and episodes the current-config baseline can't reproduce are silently dropped from pairing — skewing the pool toward PROFIT_TARGET episodes. The first accepted recommendation invalidates the next 30 days of replays. This is the era-pooling defect class this repo has been burned by twice (forecast v1/v2 pooling; the llm evidence epoch).

5. **Paper + live + IBKR pooled into one global recommendation** — `scripts/calibrate_exit_policy.py:27`. The paper book can satisfy `min_train` alone, letting a candidate clear both LCB gates purely on simulated fills while recommending a config change that governs the live book; IBKR rows (premium-priced, net==gross) mix in too. The removed per-(venue, reason) summary also removed the operator's visibility into composition.

6. **Trailing replay diverges from the runtime trigger at the boundary** — `risk/exit_calibration.py:108`. Inline `<=` where the shared helper `trailing_stop_triggered` (`exit_policy.py:80`) uses strict `>`; with `giveback=1.0` (allowed by the field bound) the replay threshold degenerates to `gross <= 0`, firing on the return-to-entry marks runtime holds through. Call the existing parameterized helper — it exists to be shared by runtime and calibration.

### High — write-path side effects

7. **A stuck exit writes one row per tick, retained 30 days** — `risk/portfolio.py:724`. Decided-but-stuck exits re-fire every 60s: terminal branches append an unsampled row per tick, and the forget loop resets the HOLD throttle each tick. One pinned dust leg ≈ 1,440 rows/day ≈ 43k retained rows — a ~10x regression vs. the old 3-day retention in exactly the stuck-exit regime, and the config comment's "one-sixth the rows" claim inverts. The dense stream also feeds `completed_episodes` with terminals banked at marks that never filled.

8. **Bare settings read on the exit path — the #420 defect class** — `risk/portfolio.py:627`. `settings.execution.exit_hold_sample_seconds` is read bare while the same function gives `exit_decision_retention_days` (line 718) the getattr-and-validate treatment for the same documented threat model. A missing field raises AttributeError outside the price try/except, discarding already-decided exits, and lands in `log.debug('exit_check.error')` — invisible to readiness: the swallowed-AttributeError signature of the 12-day outage.

9. **Hold-sample key omits exchange and coerces None-mode to paper** — `risk/portfolio.py:737`. Same market+token on two venues shares one throttle slot (one venue's sample suppresses the other's; one's exit clears the other's episode timer); on unscoped reads a live and a paper position collide under one key — the cross-book aliasing the `_peak_key` docstring's 2026-07-25 incident was written to prevent. Stale stamps from settlement/redemption-closed positions also suppress a re-entry's first HOLD, and the dict grows unboundedly.

### Medium

10. **Prune truncation + keep-first dedup dilute deltas; scale-ins mix notionals** — `risk/exit_calibration.py:152`, `:130`. The 30-day prune left-truncates straddling episodes and dedup preferentially retains those over the same identity's fully-observed re-entries (deltas collapse toward zero); per-row entry_price/size means a scale-in compares a candidate exit on the pre-scale book against a deployed exit on the post-scale book — incoherent counterfactuals.

11. **Two guard tests cannot fail** — `tests/test_exit_policy.py:394`, `tests/test_exit_calibration.py:59`. The atomic-span test's fixture row has 14 values for 13 placeholders, so the insert raises inside the contained span and the test passes with neither insert nor prune executing; the gross-vs-net trailing test's fixture marks are both below the giveback line, so gross or net passes alike. Both need discriminating fixtures.

12. **Sixth copy of the mean-LCB helper** — `risk/exit_calibration.py:137`. `backtest/ibkr_replay.py:246` already exports `mean_lcb` (further copies in ibkr_signal_study, graduation ×2, etf_calibration, ibkr_evidence). Any change to the evidence standard (e.g. small-n t-quantiles — 1.96 at n=15 is really a ~92% bound) must currently be made in six places. Import it or hoist a shared stats helper.

PR #431, opened 2026-08-08
Recover Polymarket identities from persisted stubs

DarriEy:
## Summary
- recognize condition-prefix market IDs as unresolved stubs
- retry Gamma identity recovery by the held CLOB token
- preserve the stub only when token recovery remains unavailable
- cover existing-stub recovery and settlement migration behavior

## Verification
- focused reconciliation and settlement suite: 27 passed
- complete four-PR stack: 2382 passed, 5 skipped


DarriEy (commented):

Ten finder angles, six verifier agents, gap sweep. One candidate refuted (duplicate portfolio rows via the merge path — the stale pass deletes the stub row first), two sub-claims corrected during verification (settlement double-booking is defended downstream by #423's condition-aware prior-settlement join; the silent stub retention is pre-existing silence, not a regression).

**Headline:** the retry mechanism genuinely improves on base — recovery previously never ran for the common stub path — but as built it's a **one-pass patch on a persistent-state problem**: nothing dequeues the pending stub on success, nothing removes the persisted stub row, and no test can see either. So the fix **re-seeds the duplicate-row state it was written to heal**, then re-executes the recovery (HTTP fetch + destructive REPLACE) on every pass through two 60s loops and the live exit path.

### Critical — the fix defeats itself

1. **Recovery never dequeues the stub it already queued** — `reconciler.py:124`. `_find_market_id` queues the stub before the new branch recovers the real id; `_flush_pending_stubs` then inserts the stub row anyway (`INSERT OR IGNORE` guards only the id PK; `markets` has no UNIQUE on condition_id). Every first-time recovery creates the duplicate stub+real pair the PR exists to eliminate.

2. **Recovery never converges** — `reconciler.py:407`. The docstring's "cleanup sweep" doesn't exist (zero `DELETE FROM markets` in the repo), and the un-ordered `SELECT id FROM markets WHERE condition_id=?` keeps returning the stub (discovery's `INSERT OR REPLACE` refresh gives the real row a fresh max rowid, so the stub wins the rowid scan permanently). The Gamma GET + REPLACE re-fires every reconcile pass, forever.

3. **`repair_orphaned_ids` wedges on the PK it mis-states** — `reconciler.py:521`. The renames are plain UPDATEs against the three-column PK (market_id, is_paper, **token** — the comment says two); once stub- and real-keyed rows coexist for one token, the rename raises IntegrityError, **rolls back the whole batch including the pnl_ledger settle-ref migration**, and is swallowed at `log.debug` every tick. Sticky and self-hiding. Only the pnl_ledger UPDATE got `OR IGNORE`; cost_basis/portfolio need merge/delete-stub handling.

4. **Identity flaps with per-pass Gamma availability** — `reconciler.py:124`. `recovered_id or market_id` books under the stub again on any transient Gamma failure instead of re-querying the row already ingested — dual live cost_basis/portfolio rows double-count exposure and equity into the drawdown and daily-loss gates.

### High

5. **Re-ingest clobbers enriched market rows** — `reconciler.py:386`. `INSERT OR REPLACE` rewrites a discovery-enriched row with sparse values (description/category `''`, volume/liquidity 0) on every re-fire of finding 2. Empty category **bypasses blocked-category containment** (documented at `cli/maintenance.py:76`), zeroes the correlation weight, and breaks the slug fallback.

6. **Settled-leg resurrection** — `bot.py:1522`. The settled filter keys on the market_id parsed from `settle:<id>:` refs; a leg settled while stubbed is missed once the reconciler reports the real id → the mirror re-creates the settled leg as an open position. Healing requires the repair to win a race the PK collision (finding 3) can make it lose indefinitely.

7. **Serial per-stub HTTP inside the exit path** — `reconciler.py:359`. Fresh `ClientSession` + serial GET (15s timeout) per stub-held item — including redeemables — inside `reconcile()`, which runs on two 60s loops **plus inline in every live exit attempt**. Delisted stubs retry forever (~2,880/day each); `_execute_poly_exit` pays the stall before the sell routes, then backs off 900s on a no_token. `GammaClient`'s shared session, hardened timeout, and negative cache exist and are unused here.

8. **Redeemables are recovered then dropped** — `reconciler.py:138`. `if item.redeemable: continue` runs after recovery, so redeemable stub-held positions never reach `repair_orphaned_ids` — their books stay stub-keyed forever, and the settlement sweep deactivates the stub while the sparse real row stays `active=1` (a resolved market presented as tradable).

9. **Exit token lookup stops matching for the PR's own targets** — `bot_exits.py:115`. The reconciler-sourced token join compares `rp.market_id == pos.market_id`; post-recovery rp carries the real id, the stored position the stub → miss → cost_basis fallback (stub-keyed) misses → `discovery.get_market(stub)` rejected for 0x ids → RETRYABLE loop until the repair lands (which findings 3/8 can prevent).

### Medium

10. **Same-pass split-identity for two legs of one market** — `reconciler.py:124`. The pending stub is memory-only, so leg 1 (Gamma fails) books under the stub and leg 2 (Gamma succeeds) under the real id in a single pass — an arb pairing invisible to every market_id-keyed consumer. No per-pass condition→resolution memoization.

11. **`data[0]` outside the try** — `reconciler.py:370`. A 200 with an object body (error envelope) raises KeyError(0) past the HTTP-only try, aborting reconcile mid-loop into debug-level catches. One line: `if not isinstance(data, list) or not data: return None`.

12. **Foreign-stub misdetection** — `reconciler.py:117`. Stub-ness is judged by comparing to THIS item's condition prefix, so a stub acquired via the question/ticker fallback (relistings) passes as real and the position books under a foreign market's stub — detect stub-ness of the matched row itself instead.

13. **0/0 mark floor missing** — `reconciler.py:130`. avg_price and cur_price both 0 collapse to a $0 mark with none of the flooring the trades-path defends this exact case with (lines 277-292) — understates equity into the risk gates.

14. **Out-of-order snapshot commits** — `reconciler.py:148`. reconcile() is re-entrant with no lock, and the diff stretches fetch-to-write from milliseconds to potentially minutes of serial Gamma awaits — an older pass can commit the surviving venue_positions snapshot.

15. **The new test cannot see any of this** — `tests/test_reconciler_identity.py:55`. It mocks both `_find_market_id` and `_ingest_market_from_gamma` and asserts nothing about DB state — the stub-queue, the markets table, convergence on pass 2, and the Gamma-failure leg all have zero coverage. Seed a real stub row, mock only HTTP, assert one row per condition_id after reconcile and no re-ingest on pass 2.

PR #432, opened 2026-08-08
Circuit-break failing evidence providers

DarriEy:
## Summary
- open a per-provider circuit after repeated failures
- record circuit-open fetch state instead of repeatedly paying provider timeouts
- reset provider state immediately after a successful fetch
- retry Bluesky HTTP 400 responses once without the optional freshness parameter
- preserve bounded response detail for diagnosis

## Verification
- focused provider, lineage, and redaction suite: 30 passed

## Stack
- depends on #431


DarriEy (commented):

Ten finder angles → verification (6 correctness CONFIRMED, 4 candidates REFUTED — notably the doctor log-hint blindness and the graduation-rate concern don't hold) → gap sweep. Findings 1–5, 7, 9, 11–15 are confirmed with quoted code; 6, 8, 10 are plausible with environment-dependent triggers.

**Headline:** the breaker's core state machine has a concurrency hole that can defeat the feature entirely, and the circuit's new `circuit_open` status is invisible to every downstream health consumer — so the PR can simultaneously fail to break circuits under load AND make real outages look healthier than before.

### Critical

1. **Circuit state races across concurrent gathers** — `aggregator.py:111`. State is read before the fetch await and written after; the engine runs 6-wide concurrent gathers on the shared Aggregator. Failure increments get lost (threshold counts *waves*, not failures — the breaker may never open under sustained concurrency), a slow in-flight failure can silently CLOSE a just-opened circuit, and a stale failure written after a success re-opens a circuit on a recovering provider. The new test drives sequential gathers only.

2. **Readiness self-clears during outages** — `readiness.py:300`. `check_data_sources`' fresh-outage clause needs the latest three rows all `'error'`; post-circuit rows are `'circuit_open'`, so one gather after an outage starts the criterion can never fire again and readiness reads PASS for up to 24h while the provider is dead. Widens the pre-existing `'timeout'` evasion to every outage class.

3. **`circuit_open` missing from the delivery-status failure set** — `aggregator.py:280`. A gather emptied by circuit-skipped providers records `'empty'` (the benign outcome) instead of `'partial'`; `data_quality.py`'s `failed_latest_delivery` contract then reports all-clear during a full outage of a strategy's feed.

4. **Circuit-degraded results are cached for the full TTL** — `aggregator.py:217`. The cache-store veto covers `{'timeout','error'}` but not `'circuit_open'`, contradicting the "never prolong a transient outage" invariant two lines above — a depleted snapshot taken at cooldown-second-299 serves until 359, and circuit-open (which exists only because of 3+ failures) is treated strictly *better* than the failures that caused it.

### High — the Bluesky half

5. **The 400-relaxed retry reinstates the stale-evergreen regression** — `bluesky.py:91`. Dropping `since` on AppViews that reject it removes the 10-day freshness bound, and `fetch()` has no client-side age cutoff (freshness floors at 0.2, never excludes) — a 90-day viral post outranks a 6-hour authority item and enters LLM context as current, with no log that relaxation occurred. Needs a client-side `published_at` cutoff + a log line.

6. **HTTP 400 token expiry poisons the JWT until restart** — `bluesky.py:87`. Only 401/403 clears the JWT; a PDS signaling `ExpiredToken` via 400 hits the new 400 branch, retries with the same dead token, raises — and every subsequent fetch skips createSession. The circuit then opens/re-opens on every probe: bluesky dead for days with session recreation never triggered.

7. **Doomed double round-trips forever** — `bluesky.py:115`. Nothing latches that an AppView rejects `since`, so every fetch pays a doomed request + relaxed retry for the source's lifetime (~2k extra requests/day at measured volume). Latch the rejection once the age bound is client-side.

8. **Redact-after-truncate** — `bluesky.py:99`. The body is cut to 160 chars *before* `redact_error`, inverting base.py's documented invariant — a slice can cut a credential below the redactor's matching threshold and the fragment lands in `source_fetches.error` (rendered by the dashboard). Redact first, then truncate.

### Medium

9. **Per-query 4xx rejections open the whole source** — `aggregator.py:152`. Every exception counts toward the circuit, including rejections this PR itself makes Bluesky raise — 3 queries a provider legitimately dislikes black out that source for all markets for 300s. Nothing distinguishes "provider down" from "provider disliked that query."

10. **No half-open re-arm at probe admission** — `aggregator.py:112`. At cooldown expiry every concurrent gather admits a doomed probe simultaneously (6 × 20s stalls + 6 redundant timeout rows per cycle). One line: optimistically re-arm `open_until` when admitting the probe.

11. **`ingestion_runs` records outage runs as 'ok'** — `lineage_observer.py:138`. Runs that skipped a known-dead source circuit-open aren't `'partial'`, so the evidence-audit rollup reports a healthy pipeline through a day-long outage.

12. **Duplicated breaker bookkeeping with a dead branch** — `aggregator.py:147`. The 4-line failure bookkeeping is pasted into both except handlers (timeout + error) and its `else 0.0` arm can only rewrite 0.0; a `_record_failure()` closure removes both hazards.

13. **The timeout branch has no circuit test** — `tests/test_provider_resilience.py:22`. Only the except-Exception copy is tested; a consolidation that drops the increment from the timeout path (the common failure mode, per the aggregator's own comment) ships green.

14. **Redundant `relaxed` parameter** — `bluesky.py:76`. After the 400 branch strips `since`, the `"since" in params` conjunct alone terminates the recursion on every path — the flag restates what params already encodes.

15. **WARNING per skip, silence at the transition** — `aggregator.py:119`. Dozens of identical `circuit_open` WARNINGs per cooldown across the query fan-out, while the actual threshold crossing emits no distinct event. Log one `circuit_opened` at the transition; demote the per-skip line to debug.

PR #433, opened 2026-08-08
Separate SQLite reads from the write queue

DarriEy:
## Summary
- add an independent query-only SQLite connection for non-transactional reads
- use WAL snapshot isolation so analytics reads do not block execution writes
- keep transaction-internal reads on the writer for atomic read-your-writes behavior
- close both database connections cleanly

## Verification
- focused concurrency and transaction suite: 29 passed

## Stack
- depends on #432


DarriEy (commented):
## Inline review (single reviewer — agent-fleet budget exhausted; lower coverage than the stack's earlier reviews) — promising, needs a soak gate and three fixes

The change is well-aimed: it targets exactly the measured contention (the phase-4 soak counted ~1,100+ statement/transaction queue-wait warnings per day, most of them reads queuing behind the writer), read-your-writes inside spans is correctly preserved via the `_txn_task` check, and the honest renaming of the pinned contract test shows the semantic change was understood, not stumbled into. But it changes a documented architectural invariant, so it needs the same discipline every DB change here gets:

1. **The semantic change needs a soak and a doc pass, not just a test rename** — `database.py:1611`. "Reads wait out in-flight transactions" was a *pinned contract* (the old test name) and is described in `docs/plans/db-contention-plan.md` and CLAUDE.md's architecture notes (one shared serialized connection). Post-PR, any read outside a span sees the committed snapshot and can miss a concurrently-committing batch — e.g. a risk check counting open positions while another task's entry span is mid-commit under-counts by the in-flight batch. That TOCTOU existed before (reads could schedule before the span started), but it changes from scheduling-dependent to *always* — worth quantifying in the standard 48h paper soak before the live image picks it up. Update both docs in this PR.

2. **Re-baseline the telemetry before claiming the win** — the queue-wait warning rates drop *by construction* (reads leave the queue), so post-deploy comparisons against the phase-4/5 soak baselines will overstate improvement. The contention plan's standing metric (locked-errors/day) is the one that still measures the same thing.

3. **Shutdown race on the read lane** — `database.py:270`. A fetch that passes the `self._read_db is not None` check can await a connection `close()` is concurrently closing (close() nulls it outside the serialized slot). Bound it with the existing `_closing` flag or route to the serialized fallback when closing.

4. **The hot reads that matter most still queue** — several execution-path reads use the legacy `cursor = await db.execute("SELECT ...")` style (reconciler identity lookups, parts of the exit path) and never reach `fetchone`/`fetchall`, so they still take the write lock. Worth an inventory: the PR's benefit today accrues mostly to diagnostics/strategy reads, while the head-of-line reads on the money path are the ones the motivation cites.

5. **`:memory:` databases silently keep the old semantics** — correct choice (a second `:memory:` connection would be a different database), but it means memory-based tests exercise different read semantics than production. One line in the module docstring saves the next person the surprise.

Small positives worth keeping: `PRAGMA query_only=ON` on the lane, and the new tests cover both the snapshot-read and read-your-writes legs.

PR #434, opened 2026-08-08
Report current health and classify legacy settlements

DarriEy:
## Summary
- evaluate operational errors over a bounded, configurable 30-minute window
- derive provider health from each source's latest durable fetch state
- allow successful recovery to clear historical provider failures immediately
- add schema v53 to classify blank legacy settlements as venue_unattributed
- retain blank sell attribution as an actionable lineage signal

## Verification
- focused doctor, migration, accounting, and settlement suite: 66 passed

## Stack
- depends on #433


DarriEy (commented):
## Inline review (single reviewer — agent-fleet budget exhausted) — good doctor changes, but the migration has ALREADY RUN against production in a draft form, and the merged version must account for that

### ⚠️ Deployment-state incident — must be resolved before merge

**The live database is already stamped schema 53.** The deployed image runs v52 code, which cannot have written that stamp — an out-of-band CLI invocation from this branch's checkout migrated the production DB (the `ensure_schema=False` migrate-if-behind semantics documented in #425's rollout notes). And the version that ran was NOT this PR's final text: production now shows sell rows relabeled `venue_unattributed` — the exact rows this migration's docstring says must keep `''` ("empty sell attribution remains actionable evidence of a missing entry"). Those three sells were the open leak investigation the doctor flagged on 08-08; the draft run laundered the signal into a venue-truth bucket.

Consequences and required fixes:
1. Because production reports 53, **this migration will never execute there** — any difference between the draft that ran and the merged text silently never applies.
2. **Renumber to v53→v54** and make the body corrective + complete: restore the misfiled sells (`SET strategy_source='' WHERE strategy_source='venue_unattributed' AND kind='sell'`) *and* keep the settlement relabel (idempotent no-op where the draft already applied it; still needed on any DB that never ran the draft).
3. **Add the direct migration test** the v51/v52 pattern established (DROP/set-version/call, idempotence) — absent here, and this incident is precisely the class it exists to catch.
4. Process note: no more host-side CLI runs from feature checkouts against the live file. This was the theoretical hazard in the #425 deploy notes; it is no longer theoretical.

### The doctor changes themselves — direction is right

5. **Windowed error verdict** — correct fix for the real problem observed on 08-08 (a "DEGRADED" verdict computed over a ~7-hour log tail spanning a pre-deploy incident). Liveness stays on the full tail; only recent records drive the verdict. Two nits: declare `doctor_health_window_seconds` in `MonitoringConfig` + the tracked defaults (today it exists only as a `getattr` default and the test's SimpleNamespace), and note in the detail string is already honest about the window — good.

6. **Durable provider state beats log-tail hints** — reading `source_fetches` latest-per-source fixes log-rotation blindness and is `circuit_open`-aware (which also softens the base PR's readiness gap, though readiness itself still needs #432's finding 2 addressed). Three points: the `datetime(observed_at)` wrap makes the window predicate non-sargable over a table with no retention — either index/normalize or bound the scan; the latest-timestamp join counts a source bad on same-second ties if *any* tied row is bad (conservative — fine, but comment it); and the bare `except Exception` should log the exception before degrading to "provider state unavailable" (the same fix #427 made for the lineage check).

7. **`_SOURCE_ERROR_HINTS` is now orphaned** — the constant's only consumer was the replaced check; delete it or the next reader will hunt for its purpose.

8. Label-choice note on the settlement relabel: `venue_unattributed` claims venue-truth semantics for pre-guard rows whose provenance is really "predates the guard" — downstream treatment is identical (both buckets are unjudgeable), but a sentence in the migration docstring acknowledging the era-blur would keep the provenance honest.

